PR #2, opened 2024-04-08
Improve DTS UI

DaniilKl:
Refers to issue https://github.com/Dasharo/dasharo-issues/issues/225.

### DIFF
--- a/include/dts-environment.sh
+++ b/include/dts-environment.sh
@@ -13,7 +13,7 @@ YELLOW='\033[0;33m'
 GREEN='\033[0;32m'
 BLUE='\033[0;36m'
 
-BOARD_VENDOR="$(dmidecode -s system-manufacturer)"
+SYSTEM_VENDOR="$(dmidecode -s system-manufacturer)"
 SYSTEM_MODEL="$(dmidecode -s system-product-name)"
 BOARD_MODEL="$(dmidecode -s baseboard-product-name)"
 
--- a/include/dts-environment.sh
+++ b/include/dts-environment.sh
@@ -15,8 +15,9 @@ BLUE='\033[0;36m'
 
 SYSTEM_VENDOR="$(dmidecode -s system-manufacturer)"
 SYSTEM_MODEL="$(dmidecode -s system-product-name)"
+BOARD_VENDOR="$(dmidecode -s baseboard-manufacturer)"
 BOARD_MODEL="$(dmidecode -s baseboard-product-name)"
-
+CPU_VERSION="$(dmidecode -s processor-version)"
 BIOS_VENDOR="$(dmidecode -s bios-vendor)"
 BIOS_VERSION="$(dmidecode -s bios-version)"
 DASHARO_VERSION="$(echo $BIOS_VERSION | cut -d ' ' -f 3 | tr -d 'v')"
--- a/include/dts-environment.sh
+++ b/include/dts-environment.sh
@@ -6,12 +6,26 @@
 
 # shellcheck disable=SC2034
 
-# Text Reset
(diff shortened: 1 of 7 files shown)


macpijan:
@DaniilKl I guess no new commits here yet?

scripts/ec_transition
```diff
@@ -125,7 +125,7 @@ usage() {
 
 ROOT_DIR="/"
 
-[ -z "$BOARD_VENDOR" ] && error_exit "BOARD_VENDOR not given"
+[ -z "$SYSTEM_VENDOR" ] && error_exit "BOARD_VENDOR not given"
```
macpijan, line 128: Change the help text as well?

include/dts-functions.sh
```diff
@@ -1212,3 +1224,134 @@ handle_fw_switching() {
...
+  else
+    echo -e "${BLUE}**${YELLOW}     ${DES_KEYS_OPT})${BLUE} Load your DES keys${NORMAL}"
+  fi
+  echo -e "${BLUE}*********************************************************${NORMAL}"
+  echo -e "${YELLOW}Select a menu option or${NORMAL}"
+  echo -ne "${RED}${REBOOT_OPT_UP}${NORMAL} to reboot  ${NORMAL}"
+  echo -ne "${RED}${POWEROFF_OPT_UP}${NORMAL} to poweroff  ${NORMAL}"
+  echo -e "${RED}${SHELL_OPT_UP}${NORMAL} to enter shell  ${NORMAL}"
+  if systemctl is-active sshd.socket &> /dev/null; then
+    echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to stop SHH server  ${NORMAL}"
```
macpijan, line 1353:
```suggestion
    echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to stop SSH server  ${NORMAL}"
```

include/dts-functions.sh
```diff
@@ -1212,3 +1224,134 @@ handle_fw_switching() {
...
+  else
+    echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to launch SHH server  ${NORMAL}"
```
macpijan, line 1355:
```suggestion
    echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to launch SSH server  ${NORMAL}"
```

include/dts-environment.sh
```diff
@@ -4,16 +4,32 @@
...
-BOARD_VENDOR="$(dmidecode -s system-manufacturer)"
+BLUE='\033[0;36m'
+
+# DTS options:
+HCL_REPORT_OPT="1"
+DASHARO_FIRM_OPT="2"
+REST_FIRM_OPT="3"
+DES_KEYS_OPT="4"
+REBOOT_OPT_UP="R"
+REBOOT_OPT_LOW="r"
```
macpijan, line 20: If we have to define upper and lower, we can also convert one to another: https://stackoverflow.com/questions/2264428/how-to-convert-a-string-to-lower-case-in-bash

scripts/ec_transition
DaniilKl: Amended here f921869fc12670429fe1ec65f81eb9b3253b9462.

include/dts-functions.sh
DaniilKl: Amended here dafce5b715af1efd86fd0e57035093b0188ee074.

include/dts-functions.sh
DaniilKl: Amended here dafce5b715af1efd86fd0e57035093b0188ee074

PR #7, opened 2024-04-15
include/dts-functions.sh fix error handling

WiktorG351:
adresses https://github.com/Dasharo/dasharo-issues/issues/780

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -323,7 +323,8 @@ board_config() {
           fi
           ;;
         *)
-          error_exit "Board model $SYSTEM_MODEL is currently not supported"
+          print_error "Board model $SYSTEM_MODEL is currently not supported"
+          return 1
           ;;
       esac
       ;;
@@ -410,7 +411,8 @@ board_config() {
               fi
               ;;
             *)
-              error_exit "Board model $BOARD_MODEL is currently not supported"
+              print_error "Board model $BOARD_MODEL is currently not supported"
+              return 1
               ;;
           esac
           ;;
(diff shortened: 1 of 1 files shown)


include/dts-functions.sh
```diff
@@ -703,6 +713,9 @@ check_se_creds() {
   CLOUDSEND_PASSWORD=$(sed -n '3p' < ${SE_credential_file} | tr -d '\n')
   USER_DETAILS="$CLOUDSEND_DOWNLOAD_URL:$CLOUDSEND_PASSWORD"
   board_config
+  if [ "$?" == "1" ]; then
+    return 1
+  fi
   TEST_LOGS_URL="https://cloud.3mdeb.com/index.php/s/${CLOUDSEND_LOGS_URL}/authenticate/showShare"
 
   if [ ! -v BIOS_LINK_DES ] && [ ! -v HEADS_LINK_DES ]; then
```
DaniilKl, line 721: This `error_exit` could cause the same error. Consider replacing it.

WiktorG351: what `error_exit` are You referring to? The three lines you highlighted don't contain a single `error_exit`, in fact the whole point of this PR is to get rid of them.

if you mean returning 1 here, that is precisely what needs to be done, so that [here](https://github.com/Dasharo/dts-scripts/blob/main/scripts/dts#L135-L138) (the only place in all of dts scripts where `check_se_creds` function is called) the return code can actually be utilized.

This can't cause the same error.

DaniilKl: I see I am still not filling well with GitHub GUI :smile: . I meant https://github.com/Dasharo/dts-scripts/blob/98fe8afd0951e1e4a7b8edb8b5957e8fd1d46758/include/dts-functions.sh#L722.

WiktorG351: ah i see now :smiley: 
I changed it in the [latest commit](https://github.com/Dasharo/dts-scripts/pull/7/commits/030b28ab1b972b1f5f82fceac55ed37b95649119)

PR #9, opened 2024-04-18
make exports not overwrite existing values

WiktorG351:
This is useful for tests.

If we export some mock values of for example `BIOS_VENDOR`, we do not want them to be overwritten during the test execution.

Declaring the variables like this makes it so that they are not automatically overwritten.

### DIFF
--- a/include/dts-environment.sh
+++ b/include/dts-environment.sh
@@ -7,103 +7,103 @@
 # shellcheck disable=SC2034
 
 # Text colors:
-NORMAL='\033[0m'
-RED='\033[0;31m'
-YELLOW='\033[0;33m'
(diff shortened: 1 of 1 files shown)


include/dts-environment.sh
```diff
@@ -7,103 +7,103 @@
 # shellcheck disable=SC2034
 
 # Text colors:
-NORMAL='\033[0m'
-RED='\033[0;31m'
-YELLOW='\033[0;33m'
-GREEN='\033[0;32m'
-BLUE='\033[0;36m'
+NORMAL=${NORMAL:='\033[0m'}
```
macpijan, line 10: I would limit this to the variables we intend to override only.

PR #10, opened 2024-04-22
added dts-boot for tests

WiktorG351:
This PR adds the version of `dts-boot` which creates the DES keys, which is necessary for tests on QEMU.


unit-tests/README.md
```diff
@@ -0,0 +1,268 @@
...
+We need credentials for each test variant. You can use provided template and
+fill it in accordingly.
+
+```bash
+cp des-credentials.sh.example des-credentials.sh
+```
+
+### Running automatically
+
+Some scenarios are have been already migrated into [OSFV](TBD).
```
m-iwanicki, line 26: Currently this 'link' doesn't lead anywhere. Should it stay that way?

unit-tests/dts-boot
```diff
@@ -0,0 +1,51 @@
...
+  fi
+elif [ $TEST_DES != "y" ] && [ -f "${CLOUD_PASS_FILE}" ]; then
+  rm -f "${CLOUD_PASS_FILE}"
+fi
+
+if [ -f /tmp/bios.bin ]; then
+  rm /tmp/bios.bin
+fi
+
+wget https://dl.3mdeb.com/open-source-firmware/Dasharo/novacustom_nv4x_adl/v1.7.2/novacustom_nv4x_adl_v1.7.2.rom -O /tmp/bios.bin
```
m-iwanicki, line 44: Why is this bios hardcoded? Does it matter which one is downloaded or not?

macpijan:
@WiktorG351 Shouldn't we allow for overriding **some** of the veriables? I have closed this: https://github.com/Dasharo/dts-scripts/pull/9 as it was too much. In meta-dts I have selected just a few.

unit-tests/README.md
```diff
@@ -0,0 +1,268 @@
+# DTS scripts unit tests
```
macpijan, line 1: I know it was named this way iniitally, but these are not unit tests really.

How about we at lest rename this to simply `tests` ?

unit-tests/README.md
macpijan: I think I meant to use this MR back then: https://github.com/Dasharo/open-source-firmware-validation/pull/333

macpijan:
Integrated here: https://github.com/Dasharo/dts-scripts/pull/23

PR #15, opened 2024-07-01
Check if dasharo can be installed

PLangowski:


include/dts-functions.sh
```diff
@@ -1387,3 +1387,70 @@ show_menu() {
...
+if [ $bit_30_value == 0 ]; then
+  return 1
+else
+  return 0
+fi
+}
+
+check_if_boot_guard_enabled() {
+  # Get MSR in binary
+  msr=$(rdmsr 0x13a -0 | xxd -b | awk '{for(i=2; i<=NF; i++) if($i ~ /^[01]+$/) printf "%s", $i}' | tr -d '\n')
```
miczyg1, line 1436: I gave it a quick test on my laptop. The rdmsr gives me:

```
sudo /usr/sbin/rdmsr 0x13a -0
000000130000007d
```

while this function gives:

```
+ check_if_boot_guard_enabled
++ /usr/sbin/rdmsr 0x13a -0
++ xxd -b
++ awk '{for(i=2; i<=NF; i++) if($i ~ /^[01]+$/) printf "%s", $i}'
++ tr -d '\n'
+ msr=0011000000110000001100000011000000110000001100000000000011000100110011001100000011000000110000001100000011000000110000001101110110010000001010
+ facb_fpf=0
+ verified_boot=1
+ '[' 0 == 1 ']'
+ return 1
```

The binary value of the MSR does not match the hex value from the rdmsr output. Thus the facb and verified boot values are incorrect (both should be 1 in my case.

On a platform with BtG disabled:

```
rdmsr 0x13a -0
0000000000000000
```

And the script:

```
+ check_if_boot_guard_enabled
++ rdmsr 0x13a -0
++ xxd -b
++ awk '{for(i=2; i<=NF; i++) if($i ~ /^[01]+$/) printf "%s", $i}'
++ tr -d '\n'
+ msr=0011000000110000001100000011000000110000001100000000000011000000110000001100000011000000110000001100000000000011000000110000001100000011000000001010
+ facb_fpf=0
+ verified_boot=1
+ '[' 0 == 1 ']'
+ return 1
```

miczyg1: Also why using a different converstion than here?
`hfsts6_binary=$(echo "ibase=16; obase=2; $hfsts6_value" | bc)`

miczyg1, line 1436: Have you tested what happens if this MSR does not exist? E.g. on older Intel platforms such as Minnowboard, Dell Optiplex or PC Engines apu2?

scripts/dasharo-deploy
```diff
@@ -810,6 +810,10 @@ usage() {
   echo "  $0 restore  - Restore from a previously saved backup"
 }
 
+if ! can_install_dasharo; then
```
miczyg1, line 813: This has to be used with care. The fused bit in HFSTS6 does not apply to all platforms. Dell Optiplex for example does not have the fused bit. Same with PC Engines platforms. It only applies to a subset of Intel platforms. So basically this will break deployments on many platforms that we support.

miczyg1 (changes requested):

include/dts-functions.sh
PLangowski: Should be ok now 8bb0b85a8528df8e3bbbd7815ff9772ea6c6c18d

PLangowski: Checking the exit status should be enough -  If there is no MSR it returns a non-zero status code (It was `4` on APU, but I think it's safer to catch any non-zero status) https://github.com/Dasharo/dts-scripts/pull/15/commits/8bb0b85a8528df8e3bbbd7815ff9772ea6c6c18d#diff-1e8bf7e43e9fa22b80b3469478c83c53e8e7550daed419680e53bdcbff238b82R1436

scripts/dasharo-deploy
PLangowski: Same as [here](https://github.com/Dasharo/dts-scripts/pull/15#discussion_r1671720131)

PLangowski:
@miczyg1 Could you look at the threads again and tell me if what I'm proposing is okay? And if not, can you suggest what else can be done (especially in https://github.com/Dasharo/dts-scripts/pull/15#discussion_r1671720131)?

include/dts-functions.sh
miczyg1: Not exactly. This BootGuard MSR exists on Intel platforms only, reading it on AMD should be skipped, and whole BootGuard logic as well.

scripts/dasharo-deploy
miczyg1: To not break deployments and updates we should only check for Boot Guard on platforms that we want to check for it.

include/dts-functions.sh
miczyg1: Still doesn't work:

```
+ check_if_boot_guard_enabled
+ rdmsr 0x13a -0
000000130000007d
++ rdmsr 0x13a -0
+ msr_hex=000000130000007d
++ echo 'ibase=16; obase=2; 000000130000007d'
++ bc
(standard_in) 1: syntax error
+ msr_binary=
+ binary_length=0
+ '[' 0 -lt 64 ']'
+ padding=64
++ tr ' ' 0
++ printf %64s
+ zeros=0000000000000000000000000000000000000000000000000000000000000000
+ msr_binary=0000000000000000000000000000000000000000000000000000000000000000
+ btg_capability=0
+ facb_fpf=0
+ verified_boot=0

```

PLangowski: Do we have a mechanism in DTS that checks if the platform is Intel-based? The only thing I found is the `check_intel_regions` function, but it's probably not enough.

miczyg1: You may read the `proc/cpuinfo` file. Intel processors have `vendor_id	: GenuineIntel` in the output

PLangowski: It was a matter of setting the hex string to uppercase d407c14bfb8efaccb8475b979738f1e46d6bffa6

PLangowski: d407c14bfb8efaccb8475b979738f1e46d6bffa6

scripts/dasharo-deploy
PLangowski: Now it works like this: First the script checks if the CPU vendor is Intel. If not, the other checks are skipped. If it's an Intel platform, but `rdmsr` command fails, then we assume that Boot Guard is not supported and the update runs. d407c14bfb8efaccb8475b979738f1e46d6bffa6

PLangowski:
@miczyg1 The PR is ready for another review

include/dts-functions.sh
```diff
@@ -1387,3 +1387,90 @@ show_menu() {
...
+fi
+
+bit_30_value=${hfsts6_binary:1:1}
+
+if [ $bit_30_value == 0 ]; then
+  return 1
+else
+  return 0
+fi
+}
```
miczyg1, line 1432: Indentation missing

scripts/dasharo-deploy
```diff
@@ -810,6 +810,12 @@ usage() {
   echo "  $0 restore  - Restore from a previously saved backup"
 }
 
+if check_if_intel; then
+  if ! can_install_dasharo; then
+    error_exit "Dasharo cannot be installed on this platform"
+  fi
+fi
```
miczyg1, line 817: We should run it when Dasharo is not installed. We have a fused Boot Guard on our work laptops with Dasharo.

include/dts-functions.sh
miczyg1: Now it seems to be working well.

include/dts-functions.sh
```diff
@@ -1387,3 +1387,90 @@ show_menu() {
...
+
+check_if_boot_guard_enabled() {
+  # MSR cannot be read
+  if ! rdmsr 0x13a -0; then
```
m-iwanicki, line 1436: Could you link to some documentation & page/section or explanation?
I'm trying to find those registers in

> Intel® 64 and IA-32 Architectures
Software Developer’s Manual
Volume 4:
Model-Specific Registers

PLangowski: @miczyg1 can you help here?

include/dts-functions.sh
PLangowski: https://github.com/Dasharo/dts-scripts/compare/d407c14bfb8efaccb8475b979738f1e46d6bffa6..4f681b5b519c67b505ffd19610bc8b6a4b2323cb

scripts/dasharo-deploy
PLangowski: https://github.com/Dasharo/dts-scripts/compare/d407c14bfb8efaccb8475b979738f1e46d6bffa6..4f681b5b519c67b505ffd19610bc8b6a4b2323cb

include/dts-functions.sh
pietrushnic: @miczyg1 @BeataZdunczyk this seem to be related to Dasharo TrustRoot Assesment

miczyg1: You will not find this MSR anywhere. This is BootGuard MSR

include/dts-functions.sh
```diff
@@ -1387,3 +1387,90 @@ show_menu() {
...
+  fi
+  return 1
+  }
+
+can_install_dasharo(){
+  if check_if_fused && check_if_boot_guard_enabled; then
+    return 1
+  fi
+  return 0
+}
```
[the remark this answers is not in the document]
PLangowski: https://github.com/Dasharo/dts-scripts/compare/a16bb12164ab1ac855b5430ea926cc975915bba9..3bee1318e828a7deff6c354dd2999994381f1b03

scripts/dasharo-deploy
```diff
@@ -810,6 +810,12 @@ usage() {
   echo "  $0 restore  - Restore from a previously saved backup"
 }
 
+if ! check_if_dasharo && check_if_intel; then
```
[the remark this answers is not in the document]
PLangowski: https://github.com/Dasharo/dts-scripts/compare/a16bb12164ab1ac855b5430ea926cc975915bba9..3bee1318e828a7deff6c354dd2999994381f1b03

include/dts-functions.sh
```diff
@@ -1387,3 +1387,92 @@ show_menu() {
     echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to launch SSH server  ${NORMAL}"
   fi
 }
+
+check_if_fused() {
+  file_path="/sys/class/mei/mei0/fw_status"
+
+  if [[ ! -f $file_path ]]; then
+      echo "File not found: $file_path"
+      return 2
```
miczyg1, line 1396: This indent has 6 spaces instead of 4 as in other places in this script.

include/dts-functions.sh
```diff
@@ -1387,3 +1387,92 @@ show_menu() {
...
+  fi
+
+  hfsts6_value=""
+  line_number=1
+  while IFS= read -r line; do
+      if [[ $line_number -eq 6 ]]; then
+          hfsts6_value=$line
+          break
+      fi
+      ((line_number++))
```
miczyg1, line 1406: This indent has 6 spaces instead of 4 as in other places in this script.

include/dts-functions.sh
```diff
@@ -1387,3 +1387,92 @@ show_menu() {
...
+  done < "$file_path"
+
+  if [[ -z $hfsts6_value ]]; then
+      echo "Failed to read HFSTS6 value"
+      exit 1
```
miczyg1, line 1411: This indent has 6 spaces instead of 4 as in other places in this script.

miczyg1 (changes requested):

miczyg1:
@PLangowski  please fix your editor settings. I'm done with checking the indentation already in this PR...

macpijan:
@PLangowski Can you please try to add shfmt to pre-commit?

- https://github.com/pecigonzalo/pre-commit-shfmt/blob/master/.pre-commit-hooks.yaml

include/dts-functions.sh
m-iwanicki: We could use `shfmt -i 4 include/dts-functions.sh` because shellcheck doesn't check indents.
There is also pre-commit available https://github.com/scop/pre-commit-shfmt
As this repo contains only shell scripts I think it could be very useful, though there would be a lot of fixes applied at first:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 3e92a36fc7e2..4019e6f3e9fe 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,6 +45,12 @@ repos:
       - id: markdownlint
       - id: markdownlint-fix

+  - repo: https://github.com/scop/pre-commit-shfmt
+    rev: v3.9.0-1
+    hooks:
+      - id: shfmt
+        args: ["-i", "4"]
+
   - repo: https://github.com/koalaman/shellcheck-precommit
     rev: v0.9.0
     hooks:
```

Not sure if we want to add this tool?

PLangowski: 0548c9b9045e4fb30addeb8645a0c353d11c8082

PLangowski: 0548c9b9045e4fb30addeb8645a0c353d11c8082

include/dts-functions.sh
PLangowski: 0548c9b9045e4fb30addeb8645a0c353d11c8082

include/dts-functions.sh
PLangowski: 0548c9b9045e4fb30addeb8645a0c353d11c8082

PLangowski:
I installed the tool and fixed the indents. @macpijan I will add the pre-commit config in a separate PR, because it changes a lot in the scripts.

PLangowski:
Here is the PR with shmft pre-hook https://github.com/Dasharo/dts-scripts/pull/36/files

miczyg1 (approved):

PLangowski:
Rebased

PR #16, opened 2024-07-10
Add V540TU & V560TU to board_config

m-iwanicki:
`COMPATIBLE_EC_FW_VERSION` needs to be updated when version will be known.
~~`SYSTEM_MODEL` values need to be verified (if they return `NV54xTU` and `NV56xTU`)~~


include/dts-functions.sh
```diff
@@ -322,6 +322,48 @@ board_config() {
             fi
           fi
           ;;
+          "NV54xTU")
```
mkopec, line 325: There is a slight issue in that this SMBIOS string will be the same for V540TU and V560TU. Can we add variants based on another SMBIOS string? I can set `Mainboard Version`, would that work for you?

m-iwanicki: That would be baseboard-version in dmidecode?

mkopec: Yes, that one

mkopec: Product name is `V54x_6x_TU`, baseboard version is `V540TU` and `V560TU`

m-iwanicki: Changed in [6c4ea5f](https://github.com/Dasharo/dts-scripts/pull/16/commits/6c4ea5f89a137abe1d7b137ceaeba066b1f9ad0e)

mkopec: Thanks

m-iwanicki: @mkopec Actually could you set `baseboard-product-name` instead?
That way there would be less code changes as it's already displayed for example for MSI.

mkopec: Not really, unfortunately this SMBIOS string is used for Windows Update hardware identification so it should stay the same as in vendor firmware

m-iwanicki:
@mkopec Could you confirm if config for MTL laptops is correct? Mainly this part:

```shell
PLATFORM_SIGN_KEY="customer-keys/novacustom/novacustom-open-source-firmware-release-1.x-key.asc \
                customer-keys/novacustom/dasharo-release-0.9.x-for-novacustom-signing-key.asc"
NEED_SMBIOS_MIGRATION="false"
NEED_SMMSTORE_MIGRATION="true"
NEED_BOOTSPLASH_MIGRATION="false"
NEED_BLOB_TRANSMISSION="false"
PROGRAMMER_BIOS="internal"
PROGRAMMER_EC="ite_ec:boardmismatch=force,romsize=256K,autoload=disable"
```

I'm not sure if `PLATFORM_SIGN_KEY` won't change as this key has `1.x` in name and this release will be `2.x`?

mkopec:
@m-iwanicki Two minor changes: let's enable bootsplash migration, and set the EC romsize to 128K (the EC technically has a 256K ROM size but we ship a 128K binary)

```
NEED_BOOTSPLASH_MIGRATION="true"
PROGRAMMER_EC="ite_ec:boardmismatch=force,romsize=128K,autoload=disable"
```

m-iwanicki:
@mkopec Changed in [0c69783](https://github.com/Dasharo/dts-scripts/pull/16/commits/0c69783474b08a69bf5fd1ca7262eed70c156433).
I took 256k from [docs](https://docs.dasharo.com/unified/novacustom/initial-deployment/#ec-installation). Should this be changed in the docs too?

mkopec:
Indeed it should be fixed in the docs too, thanks for pointing that out

m-iwanicki:
Fixed in [5b526e5a72050dbfc71afc86a4319027c197c00a](https://github.com/Dasharo/docs/pull/853/commits/5b526e5a72050dbfc71afc86a4319027c197c00a) in https://github.com/Dasharo/docs/pull/853

scripts/dasharo-deploy
```diff
@@ -372,15 +381,21 @@ install_ec() {
 
 install() {
   ask_for_version
-  download_artifacts
-  display_flashing_warning
-  check_flash_lock
-  verify_artifacts bios
+  if [ "$BOARD_VERSION" == "V540TU" ] || [ "$BOARD_VERSION" == "V560TU" ]; then
```
macpijan, line 384: Adding board-specific flags here in common code may be tricky in the long run.
I suspect we will have more of such models, even rather soon, e.g. the same variants with nvidia @mkopec  ?  Then we will have 4 boards here, maybe 8 in while?

Maybe add some flag indicating only EC flashing is supported, and flag is set in the board_config() ?

macpijan: But having this option of flashing EC with one click (instead of downloading EC firmware manually and flashing manually) will be nice, I hope @wessel-novacustom will find it useful :blush:

wessel-novacustom: @macpijan Very useful. We now flash EC with an own script where we ask if the platform is V540TU or V560TU. If this can somehow be automatically recognised, that would be helpful too.

mkopec: The final release firmware will have SMBIOS Baseboard Version set to V540TU or V560TU depending on the exact model. The release candidates (before rc10) don't have that yet, but EC model can also be read from DTS using `dasharo_ectool info`

macpijan (approved):

PR #18, opened 2024-07-10
dts-functions.sh: Use python3-semver in compare_versions

m-iwanicki:


include/dts-functions.sh
```diff
@@ -747,62 +747,34 @@ check_se_creds() {
...
-              fi
-
-              if [ -z "$suffix2" ] && [ ! -z "$suffix1" ]; then
-                return 1
-              fi
-          fi
-      fi
+    local ver1=
+    local ver2=
+    # convert version ending with`-rc<x>` to `-rc.<x>` where <x> is number
```
macpijan, line 770: please explain more why it is being done, maybe link semver spec as well

include/dts-functions.sh
```diff
@@ -747,62 +747,34 @@ check_se_creds() {
...
+    ver1=$(sed -r "s/-rc([0-9]+)$/-rc.\1/" <<< "$1")
+    ver2=$(sed -r "s/-rc([0-9]+)$/-rc.\1/" <<< "$2")
+    semver "$ver1" "$ver2"
+    if [ $? -eq 1 ]; then
+      return 1
     else
-      error_exit "Incorrect version format"
+      return 0;
```
macpijan, line 777:
```suggestion
      return 0
```

include/dts-functions.sh
```diff
@@ -747,62 +747,34 @@ check_se_creds() {
   fi
 }
 
+semver() {
+  # return 1 if $1 < $2, 0 if $1 = $2 or -1 if $1 > $2
+
+  local valid1=
+  local valid2=
+  valid1="$(python3 -c "import semver; print(semver.Version.is_valid('$1'))")"
```
macpijan, line 755: It can also be used as a CLI tool, maybe that's more suitable here?  `python -m semver compare/check`

m-iwanicki: Didn't know it was possible. Changed in cacba6ee9673989b3df7106cd3614261ab6f16a3
And I put everything in one function.

include/dts-functions.sh
m-iwanicki: cacba6ee9673989b3df7106cd3614261ab6f16a3

include/dts-functions.sh
m-iwanicki: Expanded comment in cacba6ee9673989b3df7106cd3614261ab6f16a3
Added information that this part is treated as alphanumeric identifier and compared like that and added link to https://semver.org

PR #25, opened 2024-08-14
wip: Add config for minnowboard platform

PLangowski:


artur-rs:
We're stopping the effort here

PR #33, opened 2024-09-23
dts-functions.sh: add NovaCustom V5x0TNx

m-iwanicki:


m-iwanicki:
@mkopec Can you check if everything looks ok? It was tested with local server and worked, config was copied from `V54x_6x_TU`

mkopec (approved):

PR #37, opened 2024-10-18
include/dts-functions.sh: Do not throw error when hfsts6 cannot be read

PLangowski:


miczyg1 (changes requested):
You have to return error, because you cannot determine if platform is fused. Thus you should not attempt installing 
Dasharo.

The proper solution would be to return when there is less that 6 HFSTS registers. But that won't completely help either, because on Bay Trail platforms like Intel MinnowMax the fuse information has to be read from TXE/ME using a message communication.

PLangowski:
@miczyg1 
If we try to install Dasharo on a fused platform it will just fail anyway, right? Is it not safe to attempt this?
If it's safe, then I think it's a better alternative to being unable to install Dasharo on an unfused platform (I'm currently experiencing this on an unfused Turbot)

miczyg1:
> @miczyg1 If we try to install Dasharo on a fused platform it will just fail anyway, right?

it will not fail, that's the problem. THe whole purpose of fuse check was to prevent bricks.

> Is it not safe to attempt this? 

it is not safe to attempt it. In most (if not all) cases you will end up with a brick.

> If it's safe, then I think it's a better alternative to being unable to install Dasharo on an unfused platform (I'm currently experiencing this on an unfused Turbot)

That's not a better alternative at all. We want to be able to deploy Dasharo on Turbot...

PR #49, opened 2024-12-03
dts-functions: add heads fw to V560TU

m-iwanicki:
@mkopec Do I need to add `HEADS_SWITCH_FLASHROM_OPT_OVERRIDE="--ifd -i bios"`?
<https://docs.dasharo.com/unified/novacustom/firmware-transition/#switching-from-dasharo-uefi-to-heads> doesn't help as it just says to use DTS.

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -348,12 +348,19 @@ board_config() {
             "V560TU")
               DASHARO_REL_NAME="novacustom_v56x_mtl"
               FLASHROM_ADD_OPT_UPDATE_OVERRIDE="--ifd -i bios"
+              HAVE_HEADS_FW="true"
+              HEADS_REL_VER_DPP="0.9.0"
+              COMPATIBLE_HEADS_EC_FW_VERSION="2024-12-20_368e08e"
+              HEADS_SWITCH_FLASHROM_OPT_OVERRIDE="--ifd -i bios"
               ;;
             *)
               print_error "Board model $BOARD_MODEL is currently not supported"
               return 1
               ;;
           esac
+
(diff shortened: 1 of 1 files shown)


PLangowski (approved):
Looks good to me

m-iwanicki:
@PLangowski can you re-review? This PR will only add support for heads for V560TU.
Additionally when switching to heads we also have to update EC

PLangowski (approved):

include/dts-functions.sh
```diff
@@ -340,6 +340,8 @@ board_config() {
           COMPATIBLE_EC_FW_VERSION="2024-07-17_4ae73b9"
           NEED_BOOTSPLASH_MIGRATION="true"
 
+          HEADS_SWITCH_FLASHROM_OPT_OVERRIDE="--ifd -i bios"
```
DaniilKl, line 343: You should add `HEADS_SWITCH_FLASHROM_OPT_OVERRIDE` only for `v560TU`.

DaniilKl (changes requested):

include/dts-functions.sh
m-iwanicki: V540TU will also need it when we add heads release for it.

m-iwanicki: https://github.com/Dasharo/dts-scripts/compare/537ce570f7939ff87902452dcf90bedffb7e2791..ba33b3f48150e066534dfe68605b6624e8b32a0e

DaniilKl: But now only v560TU needs this variable, doesn't it?

m-iwanicki: @DaniilKl I think you missed my commit by few seconds

DaniilKl (approved):

PR #50, opened 2024-12-03
dts-functions: remove ASUS

m-iwanicki:
@macpijan you said this platform is not maintained anymore? If so then there are probably more places where we should either remove or modify mentions of ASUS or DTS e.g.:

https://docs.dasharo.com/variants/asus_kgpe_d16/initial-deployment/#initial-deployment-using-dts


artur-rs:
@m-iwanicki @macpijan let's review/include this change in the next release (~v2.3.0) if possible

PR #52, opened 2024-12-10
dts-subscribtion.sh: Fix grep help message between commands

EduKav1813:
This fixes constant help menu from grep that is called between menu commands.

Grep was relying on CLOUDSEND_DOWNLOAD_URL variable, which sometimes was not set. The issue was fixed by checking for the variable first, before running the command.

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1449,9 +1449,12 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
-      local _result
+      # Return if there was an issue when asking for credentials:
+      if ! get_dpp_creds; then
+        read -p "Press Enter to continue."
+        return 0
+      fi
 
-      get_dpp_creds
 
       # Check for Dasharo Firmware for the current platform, continue to
       # packages after checking:
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1449,6 +1449,7 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
+      local _result
       # Return if there was an issue when asking for credentials:
       if ! get_dpp_creds; then
         read -p "Press Enter to continue."
--- a/include/dts-subscription.sh
+++ b/include/dts-subscription.sh
@@ -12,6 +12,11 @@ check_for_dasharo_firmware() {
 
   echo "Checking for Dasharo firmware..."
 
+  if [ -f "$DPP_CREDENTIAL_FILE" ]; then
+    print_warning "No credentials provided, cannot check for Dasharo firmware access."
+    return 1
+  fi
+
   local _check_dwn_req_resp_uefi="0"
   local _check_dwn_req_resp_heads="0"
   local _check_dwn_req_resp_seabios="0"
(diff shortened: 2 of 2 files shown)


include/dts-subscription.sh
```diff
@@ -93,7 +93,7 @@ get_dpp_creds() {
 
 login_to_dpp_server(){
   # Check if the user is already logged in, log in if not:
-  if [ -z "$(mc alias list | grep ${CLOUDSEND_DOWNLOAD_URL})" ]; then
+  if [ -n "${CLOUDSEND_DOWNLOAD_URL}" ] && [ -z "$(mc alias list | grep ${CLOUDSEND_DOWNLOAD_URL})" ]; then
```
DaniilKl, line 96: This looks like a workaround, but we need a solution. Have you checked why `CLOUDSEND_DOWNLOAD_URL`? AFAIK currently we do not have cases when some of the credentials should be empty, therefore `CLOUDSEND_DOWNLOAD_URL` should not be empty.

We have here a bigger issue that just a grep error in this line, we do not have any code that will check whether provided credentials are empty or not. If you fix only this case, there is no guarantee that another issue caused by an empty credential will not appear.

Consider following patch:

```diff
diff --git a/include/dts-functions.sh b/include/dts-functions.sh
index cbf0a2f2bc88..e70f4d99ba2e 100644
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1448,7 +1448,8 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
-      get_dpp_creds
+      # Return if there was an issue when asking for credentials:
+      get_dpp_creds || return 0
 
       # Check for Dasharo Firmware for the current platform, continue to
       # packages after checking:
diff --git a/include/dts-subscription.sh b/include/dts-subscription.sh
index 5f8505bf4639..42d6a0fdda49 100644
--- a/include/dts-subscription.sh
+++ b/include/dts-subscription.sh
@@ -12,6 +12,11 @@ check_for_dasharo_firmware() {
 
   echo "Checking for Dasharo firmware..."
 
+  if [ -f "$DPP_CREDENTIAL_FILE" ]; then
+    print_warning "No credentials provided, cannot check for Dasharo firmware access."
+    return 1
+  fi
+
   local _check_dwn_req_resp_uefi="0"
   local _check_dwn_req_resp_heads="0"
   local _check_dwn_req_resp_seabios="0"
@@ -81,6 +86,17 @@ get_dpp_creds() {
   echo ""
   read -p "Enter password:                " 'TMP_CLOUDSEND_PASSWORD'
 
+  if [ -z "$TMP_CLOUDSEND_LOGS_URL" ]; then
+    print_warning "Logs key is empty, discarting credentials..."
+    return 1
+  elif [ -z "$TMP_CLOUDSEND_DOWNLOAD_URL" ]; then
+    print_warning "Download key is empty, sicarting credentials..."
+    return 1
+  elif [ -z "$TMP_CLOUDSEND_PASSWORD" ]; then
+    print_warning "Password is empty, discarting credentials..."
+    return 1
+  fi
+
   # Export DPP creds to a file for future use. Currently these are being used
   # for both: MinIO (and its mc CLI) and cloudsend (deprecated, all DPP
   # sibscribtions will be megrated to MinIO):
```

Please, let me know what you think, and if the patch is ok and is working.

DaniilKl (changes requested):

include/dts-subscription.sh
Firminator: discarting & sicarting should be discarding

EduKav1813: @DaniilKl Added a few fixes to your patch, tested it on Qemu and it works properly.

8be200a2a87c2fb4a0e144c4dffc2bdc0bdd700d

include/dts-functions.sh
```diff
@@ -1448,7 +1448,15 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
+      # Return if there was an issue when asking for credentials:
       get_dpp_creds
+      local _dpp_creds_result=$?
+
+      if [ $_dpp_creds_result -ne 0 ]; then
```
m-iwanicki, line 1455:
```suggestion
      if ! get_dpp_creds; then
```

m-iwanicki (changes requested):

DaniilKl:
@EduKav1813, address Michał review and squash commits. Try to not add anything apart from what you change and why you change into commit message.

EduKav1813:
@m-iwanicki commited your change, sqashed all commits

include/dts-functions.sh
```diff
@@ -1449,9 +1449,12 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
-      local _result
```
DaniilKl, line 1452: Any reason you have deleted declaration of this variable?

EduKav1813: Oops, bug from rebasing on main

EduKav1813: a79f9a1c4328b3ea748b73c807f5f68a7a18a321

EduKav1813:
@DaniilKl Added fix. Instead of squashing commits again, let's just select the squash option on merge

include/dts-functions.sh
DaniilKl: You should not change anything unless you have a reason. 2561922042e60f3fbcd2aae75982a186fd245eae.

DaniilKl:
I will add your fix into v2.1.2 hotfix.

PR #55, opened 2024-12-13
Always log xtrace to file

m-iwanicki:
* /tmp/dts.log

    ```
    [01:41:55]:  [H[J[3Jc[0m
    [01:41:55]:   Dasharo Tools Suite Script 2.1.0 [0m
    [01:41:55]:  [0m (c) Dasharo <contact@dasharo.com> [0m
    [01:41:55]:  [0m Report issues at: https://github.com/Dasharo/dasharo-issues [0m
    [01:41:55]:  [0;36m*********************************************************[0m
    [01:41:55]:  [0;36m**[0m                HARDWARE INFORMATION [0m
    [01:41:55]:  [0;36m*********************************************************[0m
    ```

* /tmp/dts-verbose.log

    ```
    +[02:05:44.047]:/usr/sbin/dts-boot:29:main: source /usr/sbin/dts-environment.sh
    ++[02:05:44.048]:/usr/sbin/dts-environment.sh:9:main: source /usr/sbin/dts-hal.sh
    +++[02:05:44.049]:/usr/sbin/dts-hal.sh:25:main: source /usr/sbin/common-mock-func.sh
    ++++[02:05:44.050]:/usr/sbin/common-mock-func.sh:53:main: TEST_FLASH_LOCK=
    ++++[02:05:44.050]:/usr/sbin/common-mock-func.sh:54:main: TEST_BOARD_HAS_FD_REGION=true
    ++++[02:05:44.050]:/usr/sbin/common-mock-func.sh:55:main: TEST_BOARD_FD_REGION_RW=true
    ++++[02:05:44.051]:/usr/sbin/common-mock-func.sh:56:main: TEST_BOARD_HAS_ME_REGION=true
    ++++[02:05:44.051]:/usr/sbin/common-mock-func.sh:57:main: TEST_BOARD_ME_REGION_RW=true
    ```

When entering shell, logging to `dts.log` is disabled as redirecting output results in problems in interactive shell but verbose (xtrace) logs are still enabled just without timestamp:

```
+ echo /root/.vim/after/scripts.vim
+ shift
+ '[' 0 -ge 1 ']'
+ tail /tmp/dts-verbose.log
+ clear
+ ls
+ cp ../tmp/dts-verbose.log ../tmp/dts.log .
```

After returning to DTS UI by using `exit` normal logging is enabled again and verbose logs contain timestamp 

There is one problem, trying to enter `dts-boot` second time from shell results in error

```
/usr/sbin/dts: line 41: read: error setting terminal attributes: Interrupted system call
```

after which we end up back in shell.

@PLangowski what do you think?

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -83,8 +83,8 @@ check_if_ac() {
 fum_exit() {
     if [ "$FUM" == "fum" ]; then
       print_error "Update cannot be performed"
-      print_warning "Starting bash session - please make sure you get logs from\r
-      \r$ERR_LOG_FILE_REALPATH and $FLASHROM_LOG_FILE; then you can poweroff the platform"
+      print_warning "Starting bash session"
+      send_dts_logs ask
       /bin/bash
     fi
 }
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -150,7 +150,7 @@ it5570_shutdown() {
 }
 
 check_network_connection() {
-  if wget --spider cloud.3mdeb.com > /dev/null 2>&1; then
+  if wget --spider cloud.3mdeb.com > /dev/null 2>>"$ERR_LOG_FILE"; then
     return 0
   else
     return 1
@@ -324,7 +324,7 @@ board_config() {
           # Dasharo 0.9.0-rc10 and higher have board model in baseboard-version
           if check_if_dasharo && compare_versions "$DASHARO_VERSION" 0.9.0-rc10; then
             BOARD_MODEL="$($DMIDECODE dump_var_mock -s baseboard-version)"
-          elif ! $DASHARO_ECTOOL check_for_opensource_firm_mock info 2>/dev/null; then
+          elif ! $DASHARO_ECTOOL check_for_opensource_firm_mock info 2>>"$ERR_LOG_FILE"; then
             ask_for_model V540TU V560TU
           else
             BOARD_MODEL=$($DASHARO_ECTOOL novacustom_check_sys_model_mock info | grep "board:" |
@@ -687,22 +687,22 @@ check_flash_lock() {
 
 check_flash_chip() {
   echo "Gathering flash chip and chipset information..."
-  $FLASHROM flash_chip_name_mock -p "$PROGRAMMER_BIOS" --flash-name >> "$FLASH_INFO_FILE" 2>> "$ERR_LOG_FILE"
+  $FLASHROM flash_chip_name_mock -p "$PROGRAMMER_BIOS" --flash-name >> "$FLASH_INFO_FILE" 2>>"$ERR_LOG_FILE"
   if [ $? -eq 0 ]; then
     echo -n "Flash information: "
     tail -n1 "$FLASH_INFO_FILE"
-    FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" --flash-size 2>> /dev/null | tail -n1) / 1024 / 1024))
+    FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" --flash-size 2>>"$ERR_LOG_FILE" | tail -n1) / 1024 / 1024))
     echo -n "Flash size: "
     echo ${FLASH_CHIP_SIZE}M
   else
     for flash_name in $FLASH_CHIP_LIST
     do
-      $FLASHROM flash_chip_name_mock -p "$PROGRAMMER_BIOS" -c "$flash_name" --flash-name >> "$FLASH_INFO_FILE" 2>> "$ERR_LOG_FILE"
+      $FLASHROM flash_chip_name_mock -p "$PROGRAMMER_BIOS" -c "$flash_name" --flash-name >> "$FLASH_INFO_FILE" 2>>"$ERR_LOG_FILE"
       if [ $? -eq 0 ]; then
         echo "Chipset found"
         tail -n1 "$FLASH_INFO_FILE"
         FLASH_CHIP_SELECT="-c ${flash_name}"
-        FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} --flash-size 2>> /dev/null | tail -n1) / 1024 / 1024))
+        FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} --flash-size 2>>"$ERR_LOG_FILE" | tail -n1) / 1024 / 1024))
         echo "Chipset size"
         echo ${FLASH_CHIP_SIZE}M
         break
@@ -739,24 +739,24 @@ compare_versions() {
 
 download_bios() {
   if [ "${BIOS_LINK}" == "${BIOS_LINK_COMM}" ] || [ "${BIOS_LINK}" == "${BIOS_LINK_COMM_CAP}" ]; then
-    curl -s -L -f "$BIOS_LINK" -o $BIOS_UPDATE_FILE
+    curl -s -S -L -f "$BIOS_LINK" -o $BIOS_UPDATE_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading binary. Please
    check your internet connection"
-    curl -s -L -f "$BIOS_HASH_LINK" -o $BIOS_HASH_FILE
+    curl -s -S -L -f "$BIOS_HASH_LINK" -o $BIOS_HASH_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
    check your internet connection"
-    curl -s -L -f "$BIOS_SIGN_LINK" -o $BIOS_SIGN_FILE
+    curl -s -S -L -f "$BIOS_SIGN_LINK" -o $BIOS_SIGN_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
    check your internet connection"
   else
     USER_DETAILS="$CLOUDSEND_DOWNLOAD_URL:$CLOUDSEND_PASSWORD"
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK" -o $BIOS_UPDATE_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK" -o $BIOS_UPDATE_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL_DPP while downloading binary.
    Please check your internet connection and credentials"
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_HASH_LINK" -o $BIOS_HASH_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_HASH_LINK" -o $BIOS_HASH_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL_DPP while downloading signature.
    Please check your internet connection and credentials"
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_SIGN_LINK" -o $BIOS_SIGN_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_SIGN_LINK" -o $BIOS_SIGN_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL_DPP while downloading signature.
    Please check your internet connection and credentials"
   fi
@@ -764,23 +764,23 @@ download_bios() {
 
 download_ec() {
   if [ "${BIOS_LINK}" = "${BIOS_LINK_COMM}" ] || [ "${BIOS_LINK}" = "${BIOS_LINK_COMM_CAP}" ]; then
-    curl -s -L -f "$EC_LINK" -o "$EC_UPDATE_FILE"
+    curl -s -S -L -f "$EC_LINK" -o "$EC_UPDATE_FILE" 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading binary. Please
      check your internet connection"
-    curl -s -L -f "$EC_HASH_LINK" -o $EC_HASH_FILE
+    curl -s -S -L -f "$EC_HASH_LINK" -o $EC_HASH_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
      check your internet connection"
-    curl -s -L -f "$EC_SIGN_LINK" -o $EC_SIGN_FILE
+    curl -s -S -L -f "$EC_SIGN_LINK" -o $EC_SIGN_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
      check your internet connection"
   else
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_LINK" -o $EC_UPDATE_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_LINK" -o $EC_UPDATE_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading binary. Please
      check your internet connection and credentials"
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_HASH_LINK" -o $EC_HASH_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_HASH_LINK" -o $EC_HASH_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
      check your internet connection and credentials"
-    curl -s -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_SIGN_LINK" -o $EC_SIGN_FILE
+    curl -s -S -L -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$EC_SIGN_LINK" -o $EC_SIGN_FILE 2>>"$ERR_LOG_FILE"
     error_check "Cannot access $FW_STORE_URL while downloading signature. Please
      check your internet connection and credentials"
   fi
@@ -1001,11 +1001,11 @@ set_flashrom_update_params() {
   # We need to read whole binary (or BIOS region), otherwise cbfstool will
   # return different attributes for CBFS regions
   echo "Checking flash layout."
-  $FLASHROM read_flash_layout_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} ${FLASHROM_ADD_OPT_UPDATE} -r $BIOS_DUMP_FILE > /dev/null 2>&1
+  $FLASHROM read_flash_layout_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} ${FLASHROM_ADD_OPT_UPDATE} -r $BIOS_DUMP_FILE > /dev/null 2>>"$ERR_LOG_FILE"
   if [ $? -eq 0 ] && [ -f "$BIOS_DUMP_FILE" ]; then
     BOARD_FMAP_LAYOUT=$($CBFSTOOL layout_mock $BIOS_DUMP_FILE layout -w 2>>"$ERR_LOG_FILE")
     BINARY_FMAP_LAYOUT=$($CBFSTOOL layout_mock $1 layout -w 2>>"$ERR_LOG_FILE")
-    diff <(echo "$BOARD_FMAP_LAYOUT") <(echo "$BINARY_FMAP_LAYOUT") > /dev/null 2>&1
+    diff <(echo "$BOARD_FMAP_LAYOUT") <(echo "$BINARY_FMAP_LAYOUT") > /dev/null 2>>"$ERR_LOG_FILE"
     # If layout is identical, perform standard update using FMAP only
     if [ $? -eq 0 ]; then
       # Simply update RW_A fmap region if exists
@@ -1111,7 +1111,7 @@ handle_fw_switching() {
     done
   elif [ -n "$DPP_IS_LOGGED" ] && [ -n "$HEADS_LINK_DPP" ]; then
     local _heads_dpp=1
-    curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$HEADS_LINK_DPP" -o /dev/null
+    curl -sSfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$HEADS_LINK_DPP" -o /dev/null 2>>"$ERR_LOG_FILE"
     _heads_dpp=$?
     # We are on heads, offer switch back or perform update if DPP gives access to heads
     if [ "$DASHARO_FLAVOR" == "Dasharo (coreboot+heads)" ]; then
@@ -1316,7 +1316,7 @@ show_dpp_credentials() {
 }
 
 show_ssh_info() {
-  if systemctl is-active sshd.service &> /dev/null; then
+  if systemctl is-active sshd.service >/dev/null 2>>"$ERR_LOG_FILE"; then
     local ip=""
     ip=$(ip -br -f inet a show scope global | grep UP | awk '{ print $3 }' | tr '\n' ' ')
     # Display "check your connection" in red color in IP field in case no IPV4
@@ -1528,7 +1528,7 @@ show_footer(){
   echo -ne "${RED}${REBOOT_OPT_UP}${NORMAL} to reboot  ${NORMAL}"
   echo -ne "${RED}${POWEROFF_OPT_UP}${NORMAL} to poweroff  ${NORMAL}"
   echo -e "${RED}${SHELL_OPT_UP}${NORMAL} to enter shell  ${NORMAL}"
-  if systemctl is-active sshd.service &> /dev/null; then
+  if systemctl is-active sshd.service >/dev/null 2>>"$ERR_LOG_FILE"; then
     echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to stop SSH server  ${NORMAL}"
   else
     echo -ne "${RED}${SSH_OPT_UP}${NORMAL} to launch SSH server  ${NORMAL}"
@@ -1548,7 +1548,7 @@ footer_options(){
     "${SSH_OPT_UP}" | "${SSH_OPT_LOW}")
       wait_for_network_connection || return 0
 
-      if systemctl is-active sshd.service> /dev/null 2>&1; then
+      if systemctl is-active sshd.service >/dev/null 2>>"$ERR_LOG_FILE"; then
         print_ok "Turning off the SSH server..."
         systemctl stop sshd.service
       else
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1406,7 +1406,7 @@ main_menu_options(){
 
         if [ -n "${LOGS_SENT}" ]; then
           if ! ${CMD_DASHARO_DEPLOY} install; then
-            send_dts_logs
+            send_dts_logs ask
           fi
         fi
       else
@@ -1436,7 +1436,7 @@ main_menu_options(){
 
         # Use regular update process for everything else
         if ! ${CMD_DASHARO_DEPLOY} update; then
-          send_dts_logs
+          send_dts_logs ask
         fi
       fi
       read -p "Press Enter to continue."
@@ -1450,7 +1450,7 @@ main_menu_options(){
 
       if check_if_dasharo; then
         if ! ${CMD_DASHARO_DEPLOY} restore; then
-          send_dts_logs
+          send_dts_logs ask
         fi
       fi
       read -p "Press Enter to continue."
@@ -1593,8 +1593,14 @@ footer_options(){
   return 1
 }
 
-send_dts_logs(){
-  if [ "${SEND_LOGS_ACTIVE}" == "true" ]; then
+send_dts_logs() {
+  local send_logs="false"
+  if [ "${SEND_LOGS_ACTIVE}" = "true" ]; then
+    send_logs="true"
+  elif [ "$1" = "ask" ] && ask_for_confirmation "Do you want to send console logs to 3mdeb?"; then
+    send_logs="true"
+  fi
+  if [ "$send_logs" = "true" ]; then
     echo "Sending logs..."
 
     log_dir=$(dmidecode -s system-manufacturer)_$(dmidecode -s system-product-name)_$(dmidecode -s bios-version)
@@ -1635,6 +1641,9 @@ send_dts_logs(){
       return 1
     fi
     unset SEND_LOGS_ACTIVE
+  fi
+  if [ "$1" = "ask" ]; then
+    read -p "Press Enter to continue."
   fi
 }
 
@@ -1742,4 +1751,18 @@ check_if_intel() {
   if [ $cpu_vendor == "GenuineIntel" ]; then
     return 0
   fi
+}
+
+ask_for_confirmation() {
+  local text="$1"
+
+  read -p "$1 [N/y]: "
+  case ${REPLY} in
+    yes|y|Y|Yes|YES)
+      return 0
+      ;;
+    *)
+      return 1
+      ;;
+  esac
 }
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1415,7 +1415,7 @@ main_menu_options(){
           ${CMD_DASHARO_DEPLOY} install
           result=$?
           if [ "$result" -ne 0 ] && [ "$result" -ne 2 ]; then
-            send_dts_logs ask
+            send_dts_logs ask && return 0
           fi
         fi
       else
@@ -1447,7 +1447,7 @@ main_menu_options(){
         ${CMD_DASHARO_DEPLOY} update
         result=$?
         if [ "$result" -ne 0 ] && [ "$result" -ne 2 ]; then
-          send_dts_logs ask
+          send_dts_logs ask && return 0
         fi
       fi
       read -p "Press Enter to continue."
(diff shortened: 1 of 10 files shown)


m-iwanicki:
And I just saw that some verbose logs are duplicated (mostly in the beginning).

m-iwanicki:
And maybe write logs to permanent location so they aren't lost after reboot? This would probably require:

* some kind of log truncating/rotating so we don't run out of space. Would be good idea to implement it even when saving to `/tmp` so we don't run out of RAM (not sure if it could happen in usual use cases)
* if system is read-only (e.g. mounted in PiKVM) then maybe create tmpfs overlay so logs can still be written.

DaniilKl:
> And maybe write logs to permanent location so they aren't lost after reboot?

Unfortunately, most use cases involve iPXE. Is these cases, DTS boots from RAM and does not use any persistent storage, having logs saved when booting from USB still good thought.

m-iwanicki:
I know but it'd still be useful. Added fix for error when entering `dts-boot` multiple times. Took a while.

There is also quite a slowdown at least on QEMU in one specific part:

```

 Dasharo Tools Suite Script 2.1.1
 (c) Dasharo <contact@dasharo.com>
 Report issues at: https://github.com/Dasharo/dasharo-issues
*********************************************************
**                HARDWARE INFORMATION
*********************************************************
**    System Inf.: QEMU Standard PC (i440FX + PIIX, 1996)
** Baseboard Inf.: QEMU
**       CPU Inf.: pc-i440fx-9.1
```

And it gets stuck here for maybe 0.65 second before printing rest i.e. `**    RAM DIMM 0: Not Specified`.
It's a problem because it happens on every menu refresh. Based on verbose logs:

* starts: 

    ```
    +[17:40:51.417]:/usr/sbin/dts-functions.sh:1235:show_ram_inf: data='# dmidecode 3.5
    ```

* ends

    ```
    +[17:40:52.050]:/usr/sbin/dts-functions.sh:1273:show_ram_inf: echo -e '\033[0;36m**\033[0;33m    RAM DIMM 0: \033[0mNot Specified'
    ```

Adding stop_trace_logging/start_trace_logging at the start/end of this functions fixes this but I'm not sure if it's best way to do it, what's your opinion? I guess it's easiest to implement. Maybe we shouldn't run this function on every menu refresh but only once in e.g. `dts-boot` or use faster language (python)

```diff
diff --git a/include/dts-functions.sh b/include/dts-functions.sh
index a46d19b9f382..2b08bb9e5b3b 100644
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1230,6 +1230,7 @@ You can find more info about HCL in docs.dasharo.com/glossary\r"
 }
 
 show_ram_inf() {
+  stop_trace_logging
   # Get the data:
   local data=""
   data=$($DMIDECODE)
@@ -1272,6 +1273,7 @@ show_ram_inf() {
   for entry in "${memory_devices_array[@]}"; do
     echo -e "${BLUE}**${YELLOW}    RAM ${entry}"
   done
+  start_trace_logging "${DTS_LOG_FILE}"
 }
 
 show_header() {
```

m-iwanicki:
Slowdown is most likely due to printing timestamps

PLangowski:
If no other solution can be found for the slowdown, then I guess we can stick with your workaround.

m-iwanicki:
Added https://github.com/Dasharo/dts-scripts/pull/55/commits/58de8df1f0788d9a34fbea207ace174aa28afbba and https://github.com/Dasharo/dts-scripts/pull/55/commits/13b8a47d5b99f818fe10c9741746e285719e4f1c to deal with https://github.com/Dasharo/dasharo-issues/issues/1165.
In case of failed deployment (if dasharo-deploy error_exits) it'll ask user whether to send logs (or if automatic logging was enabled then will send logs without asking)

Disabled trace logging in `show_ram_inf` to deal with slowdown: https://github.com/Dasharo/dts-scripts/pull/55/commits/6074d44b149d8da3c9bdfb6fd475c93a1eab5e63

changed couple `2>/dev/null` to `ERR_LOG_FILE`: https://github.com/Dasharo/dts-scripts/pull/55/commits/20819bc0dece11b1f712262fceb64e0e8e781f33

Added timestamp prefix when logging to `ERR_LOG_FILE` as without it it's kinda hard to deduce where the error happened. Unfortunately I didn't manage to print script/function/line (doesn't work with `exec {fd}> >(bash_function)`, stack trace shows only where logger was started not which line called i.e. stack trace always contains only `dts-boot` and `logging` files even if called from e.g. `dts-functions.sh`): https://github.com/Dasharo/dts-scripts/pull/55/commits/7dbdccfd0b8b2cfaf020c58537dfe5e7e93711ed
Right now it's possible to compare timestamps in trace log to find which line/command does the error belong to.

I guess PR is ready for review, we could maybe move console logs to `/var/local/` but we would probably need to enable truncating those files (or maybe rotating but that'd need to be done during system boot, before `dts-boot` starts as after that file descriptor for normal console logs are kept for the whole `dts-boot` lifetime and trace logs only closes fd when stopping trace logging (only done in `show_ram_inf` for split second) which would result in logs being written to old files instead of new ones).

PLangowski:
When using qemu with telnet I have 2 files for each log:

```
dts-verbose_tty1.log   dts_tty1.log
dts-verbose_ttyS0.log  dts_ttyS0.log
```

m-iwanicki:
Each tty starts it's own dts-boot (serial, graphical QEMU, ssh), that way we don't mix logs from different tty.
I could do this another way: add tty prefix to logs. Currently I only send logs from tty that requested those logs to be sent.

scripts/dts-boot
```diff
@@ -5,15 +5,32 @@
...
+export BASH_ENV="$SBIN_DIR/logging"
+export ERR_LOG_FILE_REALPATH
+export DTS_LOG_FILE
+export DTS_VERBOSE_LOG_FILE
+export ERR_LOG_FILE
+export SHELLOPTS
+
+ERR_LOG_FILE_REALPATH="/var/local/dts-err_$(basename "$(tty)").log"
+DTS_LOG_FILE="/tmp/dts_$(basename "$(tty)").log"
+DTS_VERBOSE_LOG_FILE="/tmp/dts-verbose_$(basename "$(tty)").log"
```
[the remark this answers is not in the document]
DaniilKl: you could also explain why you have added `ERR_LOG_FILE_REALPATH` variable, as it is not clear at first glance.

DaniilKl (changes requested):

include/dts-functions.sh
```diff
@@ -1429,7 +1434,9 @@ main_menu_options(){
         fi
 
         # Use regular update process for everything else
-        ${CMD_DASHARO_DEPLOY} update
+        if ! ${CMD_DASHARO_DEPLOY} update; then
```
DaniilKl, line 1437: `dasharo_deploy update` will return `1` in case of platform does not have updates available, and will trigger your `send_dts_logs ask`:

```bash
Enter an option:
2
Waiting for network connection ...
Network connection have been established!
Checking if board is Dasharo compatible.
Waiting for system clock to be synced ...
Current Dasharo version: 0.2.0
Latest available Dasharo version for your subscribtion: 0.2.0
No update available for your machine
Do you want to send console logs to 3mdeb? [N/y]: 
```

This is counterintuitive, could you fix it?

include/dts-functions.sh
```diff
@@ -1230,6 +1230,8 @@ You can find more info about HCL in docs.dasharo.com/glossary\r"
 }
 
 show_ram_inf() {
+  # trace logging is quite slow due to timestamp (calls 'date')
```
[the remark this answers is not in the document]
m-iwanicki: It causes those delays because it's external program called for each line in subshell.
Each refresh generates about 800 logged lines of which about 660 is a result of this function (mostly because of parsing long output in loop).

scripts/dts-boot
```diff
@@ -5,15 +5,32 @@
 # SPDX-License-Identifier: Apache-2.0
 
 SBIN_DIR="/usr/sbin"
-BIN_DIR="/usr/bin"
```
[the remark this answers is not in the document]
m-iwanicki: ? I did, that was the only reference

include/dts-functions.sh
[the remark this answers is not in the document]
DaniilKl: Could you make a comment in the function using your response?

include/dts-functions.sh
```diff
@@ -888,7 +888,7 @@ check_blobs_in_binary() {
 
   # If there is no descriptor, there is no ME as well, so skip the check
   if [ $BOARD_HAS_FD_REGION -ne 0 ]; then
-    ME_OFFSET=$($IFDTOOL check_blobs_in_binary_mock -d $1 2> /dev/null | grep "Flash Region 2 (Intel ME):" | sed 's/Flash Region 2 (Intel ME)\://' | awk '{print $1;}')
+    ME_OFFSET=$($IFDTOOL check_blobs_in_binary_mock -d $1 2>>"$ERR_LOG_FILE" | grep "Flash Region 2 (Intel ME):" | sed 's/Flash Region 2 (Intel ME)\://' | awk '{print $1;}')
```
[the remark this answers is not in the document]
m-iwanicki: A lot of those are stdout to /dev/null or to completely get rid of output e.g. in `if`'s where error is expected.
Some commands always output to stderr e.g. dd or flashrom:

```sh
bash-5.2# flashrom -p internal >/dev/null
Warning: Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.
New value is 0x8b.
```

Which ones do you want redirected? Remember that the only way to connect those errors to lines is searching timestamps in trace logs (which won't be 1-to-1)

include/dts-functions.sh
```diff
@@ -1591,9 +1595,10 @@ send_dts_logs(){
 
     mkdir $log_dir
     cp ${DTS_LOG_FILE} $log_dir
+    cp ${DTS_VERBOSE_LOG_FILE} $log_dir
 
-    if [ -f ${ERR_LOG_FILE} ]; then
-      cp ${ERR_LOG_FILE} $log_dir
+    if [ -f ${ERR_LOG_FILE_REALPATH} ]; then
+      cp ${ERR_LOG_FILE_REALPATH} $log_dir
     fi
```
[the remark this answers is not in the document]
m-iwanicki: `cp /tmp/a/b/c/file some/dir` would result in `some/dir/file` being created

include/dts-functions.sh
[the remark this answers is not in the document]
DaniilKl: Logs from every command might be useful during debugging. For example these commands send their STDERR everywhere but error log file:

```bash
include/hal/dts-hal.sh:147:  $DASHARO_ECTOOL check_for_opensource_firm_mock info > /dev/null 2>&1
include/hal/dts-hal.sh:193:  _mode="$($SETPCI check_me_op_mode_mock -s 00:16.0 42.B 2> /dev/null | cut -c2-)"
include/dts-functions.sh:153:  if wget --spider cloud.3mdeb.com > /dev/null 2>&1; then
include/dts-functions.sh:326:          elif ! $DASHARO_ECTOOL check_for_opensource_firm_mock info 2>/dev/null; then
include/dts-functions.sh:693:    FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" --flash-size 2>> /dev/null | tail -n1) / 1024 / 1024))
include/dts-functions.sh:704:        FLASH_CHIP_SIZE=$(($($FLASHROM flash_chip_size_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} --flash-size 2>> /dev/null | tail -n1) / 1024 / 1024))
include/dts-functions.sh:1003:  $FLASHROM read_flash_layout_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} ${FLASHROM_ADD_OPT_UPDATE} -r $BIOS_DUMP_FILE > /dev/null 2>&1
include/dts-functions.sh:1316:  if systemctl is-active sshd.service &> /dev/null; then
include/dts-functions.sh:1506:  if systemctl is-active sshd.service &> /dev/null; then
include/dts-functions.sh:1526:      if systemctl is-active sshd.service> /dev/null 2>&1; then
reports/dasharo-hcl-report:222:if pushd logs/acpi &> /dev/null; then
reports/dasharo-hcl-report:225:  popd &> /dev/null || return 1
scripts/dasharo-deploy:520:      dd if=/tmp/rom.bin of=/tmp/romhole.bin skip=$((0x17C0000)) bs=128K count=1 iflag=skip_bytes > /dev/null 2>&1
scripts/dasharo-deploy:604:    $FLASHROM read_firm_mock -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} ${FLASHROM_ADD_OPT_READ} -r $BIOS_DUMP_FILE > /dev/null 2>/dev/null
scripts/dasharo-deploy:607:      diff <(echo "$BINARY_KEYS") <(echo "$FLASH_KEYS") > /dev/null 2>&1
```

Maybe you could as well dump some info from these curls:

```bash
include/dts-subscription.sh:42:      _check_dwn_req_resp_uefi=$(curl -L -I -s -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP" -o /dev/null -w "%{http_code}")
include/dts-subscription.sh:46:      _check_dwn_req_resp_uefi_cap=$(curl -L -I -s -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP_CAP" -o /dev/null -w "%{http_code}")
include/dts-subscription.sh:50:      _check_dwn_req_resp_heads=$(curl -L -I -s -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$HEADS_LINK_DPP" -o /dev/null -w "%{http_code}")
include/dts-subscription.sh:54:      _check_dwn_req_resp_seabios=$(curl -L -I -s -f -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP_SEABIOS" -o /dev/null -w "%{http_code}")
include/dts-subscription.sh:57:    _check_logs_req_resp=$(curl -L -I -s -f -H "$CLOUD_REQUEST" "$TEST_LOGS_URL" -o /dev/null -w "%{http_code}")
include/dts-functions.sh:1113:    curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$HEADS_LINK_DPP" -o /dev/null
scripts/dasharo-deploy:83:      curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP" -o /dev/null
scripts/dasharo-deploy:91:      curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP_CAP" -o /dev/null
scripts/dasharo-deploy:99:      curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$BIOS_LINK_DPP_SEABIOS" -o /dev/null
scripts/dasharo-deploy:107:      curl -sfI -u "$USER_DETAILS" -H "$CLOUD_REQUEST" "$HEADS_LINK_DPP" -o /dev/null
```

scripts/dts-boot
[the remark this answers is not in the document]
m-iwanicki: 56fd2d6b776299742a5a44341d16cb2fe89b5505

Not sure about `BASH_ENV` as it's sourced by every script we call. I would rather not assign anything else there as if for some reason we wanted to change variable value during runtime then it would only persist until we enter another script.
I mostly used it to set PS4 to be the same everywhere otherwise it's not set (even if exported) in subscripts called as command e.g. `$SBIN_DIR/dts`

scripts/logging
```diff
@@ -0,0 +1,46 @@
...
+        echo "[$(date +%T.%3N)]: $LINE"
+    done
+}
+
+start_logging() {
+    local file="$DTS_LOG_FILE"
+    if [ -z "$LOGGING" ]; then
+        LOGGING="Y"
+        if [ -z "$DESC" ]; then
+            exec {DESC}> >(tee >(stdbuf -i0 -oL -eL ts "[%T]: " >> "$file"))
```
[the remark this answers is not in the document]
m-iwanicki: 11c50486d910707de8babe7bf9591e034207bd87

scripts/logging
```diff
@@ -0,0 +1,46 @@
...
+    while read -r LINE
+    do
+        echo "[$(date +%T.%3N)]: $LINE"
+    done
+}
+
+start_logging() {
+    local file="$DTS_LOG_FILE"
+    if [ -z "$LOGGING" ]; then
+        LOGGING="Y"
```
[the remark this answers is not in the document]
m-iwanicki: 11c50486d910707de8babe7bf9591e034207bd87

include/dts-functions.sh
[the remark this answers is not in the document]
m-iwanicki: 965855f749476003b9040eb7a73a2a14725d52eb

We don't need to disable trace logging but temporarily change PS4 variable value so it doesn't contain call to date but I'm not sure if we want 650 lines of spam for every screen refresh (which increases log size by 60 KB for each refresh instead of just around 7 KB)

include/dts-functions.sh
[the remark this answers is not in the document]
m-iwanicki: As I said, dd always prints to stderr (whether there is error or not)
Do you want each curl output in error logs? Should I create new ones? Add to `DTS_LOG_FILE`? If to `DTS_LOG_FILE` then should I also add timestamps there?

include/dts-functions.sh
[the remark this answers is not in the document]
m-iwanicki: The final name will be the same as always e.g:
`QEMU_Standard_PC_(i440FX_+_PIIX,_1996)_1.16.3-3.fc41_754df0b9-bf8e-5beb-bc03-95d22bfae38c_2024_12_20_09_31_28_626870134.tar.gz`.

We are not using absolute path anywhere and even if we did then command would be e.g. `tar czf /tmp/log.tar.gz /tmp/log` which would create `log.tar.gz` file in tmp directory but:

* It's possible the command would fail because we would be trying to create file in folder we are trying to archive (recursion)
* files in archive itself would contain this path e.g. `tmp/log/...` (without leading `/`).

In other words it works ok

include/dts-functions.sh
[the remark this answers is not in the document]
m-iwanicki: I was asking if you want curl output (`-o /dev/null` == `>/dev/null`) as in:

```
curl -L -I -f --silent -u 'abc:abc' 'https://cloud.3mdeb.com/public.php/webdav' -X PROPFIND -H "Depth: 5"
HTTP/1.1 401 Unauthorized
Server: nginx/1.22.1
Date: Fri, 20 Dec 2024 11:12:38 GMT
Content-Type: application/xml; charset=utf-8
Content-Length: 234
Connection: keep-alive
Strict-Transport-Security: max-age=15552000; includeSubDomains
Referrer-Policy: no-referrer
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
X-Robots-Tag: noindex, nofollow
X-XSS-Protection: 1; mode=block
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
Set-Cookie: 
 path=/; secure; HttpOnly; SameSite=Lax
Content-Security-Policy: default-src 'none';
Set-Cookie: __Host-nc_sameSiteCookielax=true; path=/; httponly;secure; expires=Fri, 31-Dec-2100 23:59:59 GMT; SameSite=lax
Set-Cookie: __Host-nc_sameSiteCookiestrict=true; path=/; httponly;secure; expires=Fri, 31-Dec-2100 23:59:59 GMT; SameSite=strict
WWW-Authenticate: Basic realm="3mdeb", charset="UTF-8"
Strict-Transport-Security: max-age=31536000; includeSubDomains
```

It'll output this result whether it succeeds or not. Or do you want only errors?
As we are using `--silent` flag curl doesn't return errors. I guess I could also add `-S` flag so we can redirect:

```
curl: (22) The requested URL returned error: 401
```

Added `-S` flag and error redirection to curls: 104527a155d135b42b2a2c6fe775c0057a50967b

scripts/dts-boot
```diff
@@ -5,15 +5,32 @@
...
+ERR_LOG_FILE_REALPATH="/var/local/dts-err_$(basename "$(tty)").log"
+DTS_LOG_FILE="/tmp/dts_$(basename "$(tty)").log"
+DTS_VERBOSE_LOG_FILE="/tmp/dts-verbose_$(basename "$(tty)").log"
+
+# shellcheck source=./logging
+source "$BASH_ENV"
+start_trace_logging
+start_logging
+if [ -z "$ERR_LOG_FILE" ]; then
+    exec {ERR_LOG_FILE}> >(logger >> "$ERR_LOG_FILE_REALPATH")
```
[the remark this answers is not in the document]
m-iwanicki: 104527a155d135b42b2a2c6fe775c0057a50967b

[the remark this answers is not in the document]
DaniilKl: @m-iwanicki, wrong commit?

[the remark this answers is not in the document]
m-iwanicki: https://github.com/Dasharo/dts-scripts/pull/55/commits/a90f235e7766c0b0ff4f18cf7abcff99da74afca
Sometimes ctrl+c doesn't catch

include/dts-functions.sh
m-iwanicki: 85d646c6d5eeeb96a686f69a597077bd1f27a06a

Whole PR will need to be thoroughly tested

m-iwanicki: force-pushed: 47c8d385650026bd0c82d4cfc66f1dbf55c16841, quick patch job, should probably think of something better. Maybe you have some ideas? If it works then I guess it could stay as is an be either fixed later on be made obsolete when we migrate to GO.

include/dts-functions.sh
```diff
@@ -1722,21 +1741,16 @@ check_if_intel() {
...
-    unset VERBOSE_ACTIVE
-    # Remove the -x option
-    CMD_DASHARO_DEPLOY=$(echo $CMD_DASHARO_DEPLOY | sed 's|^/usr/bin/env bash -x ||')
-    CMD_DASHARO_HCL_REPORT=$(echo $CMD_DASHARO_HCL_REPORT | sed 's|^/usr/bin/env bash -x ||')
-    CMD_EC_TRANSITION=$(echo $CMD_EC_TRANSITION | sed 's|^/usr/bin/env bash -x ||')
-    CMD_CLOUD_LIST=$(echo $CMD_CLOUD_LIST | sed 's|^/usr/bin/env bash -x ||')
-    set +x
-  fi
+ask_for_confirmation() {
+  local text="$1"
```
PLangowski, line 1745: This variable is not used

include/dts-functions.sh
```diff
@@ -1207,7 +1211,7 @@ handle_fw_switching() {
 
 sync_clocks() {
   echo "Waiting for system clock to be synced ..."
-  chronyc waitsync 10 0 0 5 &> /dev/null
+  chronyc waitsync 10 0 0 5 >/dev/null 2>>ERR_LOG_FILE
```
PLangowski, line 1214:
```suggestion
  chronyc waitsync 10 0 0 5 >/dev/null 2>>$ERR_LOG_FILE
```

PLangowski (changes requested):

include/dts-functions.sh
m-iwanicki: 1ec8742b5602d8d26800875309ab20d511f5dffe

include/dts-functions.sh
m-iwanicki: 1ec8742b5602d8d26800875309ab20d511f5dffe

PLangowski (commented):
The logging works as expected. However, when I tried to update MSI PRO Z690-A DDR5, it failed:
```
+[08:35:34.758]:/usr/sbin/dts-hal.sh:136:tool_wrapper: flashrom -p internal --ifd -i bios -i fd -w /tmp/biosupdate
+++[08:35:34.763]:/usr/sbin/logging:6:logger: date +%T.%3N
++[08:35:34.765]:/usr/sbin/logging:6:logger: echo '[08:35:34.764]: Warning: Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.'
++[08:35:34.766]:/usr/sbin/logging:4:logger: read -r LINE
+++[08:35:34.767]:/usr/sbin/logging:6:logger: date +%T.%3N
++[08:35:34.768]:/usr/sbin/logging:6:logger: echo '[08:35:34.767]: New value is 0x8b.'
++[08:35:34.769]:/usr/sbin/logging:4:logger: read -r LINE
+++[08:37:43.209]:/usr/sbin/logging:6:logger: date +%T.%3N
++[08:37:43.210]:/usr/sbin/logging:6:logger: echo '[08:37:43.209]: FAILED at 0x00209046! Expected=0xff, Found=0x00, failed byte count from 0x00000000-0x01ffffff: 0x153'
+[08:37:43.211]:/usr/sbin/dts-hal.sh:138:tool_wrapper: return 3
++[08:37:43.211]:/usr/sbin/logging:4:logger: read -r LINE
+++[08:37:43.211]:/usr/sbin/logging:6:logger: date +%T.%3N
+[08:37:43.211]:/usr/sbin/dasharo-deploy:828:deploy_firmware: error_check 'Failed to update Dasharo firmware'
+[08:37:43.212]:/usr/sbin/dts-functions.sh:107:error_check: _error_code=3
++[08:37:43.212]:/usr/sbin/logging:6:logger: echo '[08:37:43.212]: Your flash chip is in an unknown state.'
+[08:37:43.213]:/usr/sbin/dts-functions.sh:108:error_check: _error_msg='Failed to update Dasharo firmware'
```
I don't know if this failure is connected with the changes in this PR, but it's worth investigating

m-iwanicki:
That's flashrom error:

```
++[08:37:43.210]:/usr/sbin/logging:6:logger: echo '[08:37:43.209]: FAILED at 0x00209046! Expected=0xff, Found=0x00, failed byte count from 0x00000000-0x01ffffff: 0x153'
```

Probably during verification. Also had the same problem on current release first try failed and second worked.

I see now that xtrace prints:

```
++[08:37:43.211]:/usr/sbin/logging:4:logger: read -r LINE
+++[08:37:43.211]:/usr/sbin/logging:6:logger: date +%T.%3N
```

when using `>>$ERR_LOG_FILE`. Adds some noise to the logs but I'm not sure if it's worth complicating logging even more just to remove this line from being logged.

PLangowski (commented):
Could you maybe put the log files in a directory inside `/tmp`, like `/tmp/logs`?

m-iwanicki:
@PLangowski a75cdb01d9df

PLangowski (approved):

DaniilKl:
>  Probably during verification. Also had the same problem on current release first try failed and second worked.

Has this issue been reported or noted somewhere? Have you maybe checked why exactly this happened? Maybe we should report it to Dasharo team?

DaniilKl:
I have resolved my threads, but I need to check some places again, and verify it on some platform or QEMU. Will do ir tomorrow.

m-iwanicki:
> Has this issue been reported or noted somewhere

Yes, in our internal ticket to fix bricked platform. I sent you link on MM.

include/dts-functions.sh
```diff
@@ -1429,7 +1443,11 @@ main_menu_options(){
         fi
 
         # Use regular update process for everything else
-        ${CMD_DASHARO_DEPLOY} update
+        ${CMD_DASHARO_DEPLOY} update;
```
DaniilKl, line 1446: Why `;`?

include/dts-functions.sh
```diff
@@ -1560,20 +1576,19 @@ footer_options(){
...
 }
 
-send_dts_logs(){
-  if [ "${SEND_LOGS_ACTIVE}" == "true" ]; then
+send_dts_logs() {
+  local send_logs="false"
+  if [ "${SEND_LOGS_ACTIVE}" = "true" ]; then
+    send_logs="true"
+  elif [ "$1" = "ask" ] && ask_for_confirmation "Do you want to send console logs to 3mdeb?"; then
+    send_logs="true"
```
DaniilKl, line 1589: Why not to set `SEND_LOGS_ACTIVE` to `true` here?

include/dts-functions.sh
```diff
@@ -1591,15 +1606,16 @@ send_dts_logs(){
 
     mkdir $log_dir
     cp ${DTS_LOG_FILE} $log_dir
+    cp ${DTS_VERBOSE_LOG_FILE} $log_dir
```
DaniilKl, line 1609: Could `log_dir` be inside `TMP_LOG_DIR` so to keep everything in one place?

include/dts-functions.sh
```diff
@@ -1722,21 +1741,16 @@ check_if_intel() {
...
-  fi
+ask_for_confirmation() {
+  local text="$1"
+
+  read -p "$text [N/y]: "
+  case ${REPLY} in
+    yes|y|Y|Yes|YES)
+      return 0
+      ;;
+    *)
```
DaniilKl, line 1752: You should check for `No` explicitly. User might misclick some another character that will cause logs absence. Actually, we should write a separate function for all inputs from users to standardize it in entire DTS, but it is a task for another PR.

DaniilKl (changes requested):

include/dts-functions.sh
m-iwanicki: because SEND_LOGS_ACTIVE is exported global variable which I don't want to change

include/dts-functions.sh
m-iwanicki: e5b639ea6cd9

include/dts-functions.sh
m-iwanicki: e5b639ea6cd9

include/dts-functions.sh
m-iwanicki: Forgot to remove when changing if in https://github.com/Dasharo/dts-scripts/commit/47c8d385650026bd0c82d4cfc66f1dbf55c16841#diff-1e8bf7e43e9fa22b80b3469478c83c53e8e7550daed419680e53bdcbff238b82R1446:

```diff
-        if ! ${CMD_DASHARO_DEPLOY} update; then
+        ${CMD_DASHARO_DEPLOY} update;
```

e5b639ea6cd9

m-iwanicki:
@DaniilKl rebased

DaniilKl (approved):

scripts/logging
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+logger() {
+    while read -r LINE
+    do
+        echo "[$(date +%T.%3N)]: $LINE"
+    done
+}
+
+start_logging() {
```
DaniilKl, line 10: Noted some strange behavior while testing.

1) Enable logging by pressing `L`:

```bash
*********************************************************
R to reboot  P to poweroff  S to enter shell  
K to launch SSH server  L to disable sending DTS logs 

Enter an option:
```

2) Go to Shell and see logs being sent:

```bash
Entering shell, to leave type exit and press Enter or press LCtrl+D

Sending logs...
Tavinus Cloud Sender v2.2.8

> Using password from environment

SENDING SINGLE FILE
===================

Emulation_QEMU_x86_q35_ich9_Dasharo_(coreboot+UEFI)_v0.2.0_91381668-382c-51ad-bfe7-d11c3a225845_2025_01_10_10_19_30_263919342.tar.gz > 
######################################################################## 100.0%

SUMMARY
=======

 > All Curl calls exited without errors and no WebDAV errors were detected
 > Attempt to send completed > Emulation_QEMU_x86_q35_ich9_Dasharo_(coreboot+UEFI)_v0.2.0_91381668-382c-51ad-bfe7-d11c3a225845_2025_01_10_10_19_30_263919342.tar.gz
bash-5.2# 
```

3) Return to DTS menu by pressing `Ctrl+d` and see that sending logs should be enabled again:

```bash
*********************************************************
R to reboot  P to poweroff  S to enter shell  
K to launch SSH server  L to enable sending DTS logs 

Enter an option:
```

Is it supposed behavior? If so, why?

m-iwanicki: That's how it was implemented by @PLangowski, after logs are sent, sending logs is disabled. I think we should probably keep it enabled. What do you think?

include/dts-functions.sh
```diff
@@ -1403,6 +1413,10 @@ main_menu_options(){
 
         if [ -n "${LOGS_SENT}" ]; then
           ${CMD_DASHARO_DEPLOY} install
+          result=$?
+          if [ "$result" -ne 0 ] && [ "$result" -ne 2 ]; then
+            send_dts_logs ask
```
DaniilKl, line 1418:
```suggestion
            send_dts_logs ask && return 0
```

So not to see `Press Enter to continue.` twice.

include/dts-functions.sh
```diff
@@ -1431,6 +1445,10 @@ main_menu_options(){
 
         # Use regular update process for everything else
         ${CMD_DASHARO_DEPLOY} update
+        result=$?
+        if [ "$result" -ne 0 ] && [ "$result" -ne 2 ]; then
+          send_dts_logs ask
```
DaniilKl, line 1450:
```suggestion
          send_dts_logs ask return 0
```

So not to see `Press Enter to continue.` twice.

include/dts-functions.sh
```diff
@@ -1442,7 +1460,9 @@ main_menu_options(){
       [ "${SYSTEM_VENDOR}" = "QEMU" ] || [ "${SYSTEM_VENDOR}" = "Emulation" ] && return 0
 
       if check_if_dasharo; then
-        ${CMD_DASHARO_DEPLOY} restore
+        if ! ${CMD_DASHARO_DEPLOY} restore; then
+          send_dts_logs ask
```
DaniilKl, line 1464:
```suggestion
          send_dts_logs ask return 0
```

So not to see `Press Enter to continue.` twice.

DaniilKl (changes requested):

scripts/logging
DaniilKl: I agree, we should keep it enabled. User might not notice this change, and it could potentially cause logs absence in case of some other issue.

include/dts-functions.sh
m-iwanicki: d394c656fe39

include/dts-functions.sh
m-iwanicki: d394c656fe39

scripts/logging
m-iwanicki: d447a1711a11

DaniilKl:
Actually, why are you sending logs before entering shell, rebooting or poweroff? IIRC there is some thread explaining this but I cannot find it. Could you leave a comment there explaining the reasson to do so?

m-iwanicki:
@DaniilKl I don't, those changes were already there but we are sending logs before reboot/poweroff because otherwise those would be lost. I'm not sure why when entering shell. @PLangowski could you explain?

PLangowski:
@m-iwanicki That was just an initial idea that went through some time ago. The idea was that upon entering the shell the script was no longer in control, so it couldn't send the logs at any later moment, but I'm not sure if that's even correct.

m-iwanicki:
Depends what you do in script, if you call `reboot/shutdown` then yes, it can't, if you `exit`, `CTRL+D` or `dts-boot` then log sending would continue to be enabled

DaniilKl:
> Depends what you do in shell, if you call reboot/shutdown then yes

I see. Ok then.

DaniilKl (approved):

m-iwanicki:
I'm not sure that we actually need to send logs before entering shell (except in FUM which enters it automatically on error). I guess we could alias reboot/shutdown to sent logs:

```sh
alias reboot="send_dts_logs; reboot"
```

but I'm not sure if it's good idea.

But I'm not sure if it's

PR #57, opened 2024-12-18
fix DTS extensions message

DaniilKl:
Related PR:
https://github.com/Dasharo/meta-dts/pull/208

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1449,22 +1449,40 @@ main_menu_options(){
       return 0
       ;;
     "${DPP_KEYS_OPT}")
+      local _result
+
       get_dpp_creds
 
       # Check for Dasharo Firmware for the current platform, continue to
       # packages after checking:
       check_for_dasharo_firmware
+      _result=$?
+
+      echo "Your credentials give access to:"
+      echo -n "Dasharo Pro Package (DPP): "
+
+      if [ $_result -eq 0 ]; then
+        # FIXME: what if credentials have access to
+        # firmware, but check_for_dasharo_firmware will not detect any platform?
+        # According to check_for_dasharo_firmware it will return 1 in both
+        # cases which means that we cannot detect such case.
+        print_ok "YES"
+      else
+        echo "NO"
+      fi
+
+      echo -n "DTS Extensions: "
 
       # Try to log in using available DPP credentials, start loop over if login
       # was not successful:
       login_to_dpp_server
       if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


include/dts-functions.sh
```diff
@@ -1453,19 +1453,34 @@ main_menu_options(){
...
+        # should be inside condition. FIXME: what if credentials have access to
+        # firmware, but check_for_dasharo_firmware will not detect any platform?
+        # According to check_for_dasharo_firmware it will return 1 in both
+        # cases which means that we cannot detect such case.
+        echo "Your credentials give access to:"
+        echo -n "Dasharo Pro Package (DPP): "
+        print_ok "YES"
+      else
+        echo "Your credentials give access to:"
+	echo -n "Dasharo Pro Package (DPP): "
```
EduKav1813, line 1467: You can move out the common part, since you will print those messages regardless of the result of the function

DaniilKl: Check comments above.

include/dts-functions.sh
```diff
@@ -1453,19 +1453,34 @@ main_menu_options(){
...
       # Check for Dasharo Firmware for the current platform, continue to
       # packages after checking:
-      check_for_dasharo_firmware
+      if check_for_dasharo_firmware; then
+        # check_for_dasharo_firmware prints some messages, so these echoes
+        # should be inside condition. FIXME: what if credentials have access to
+        # firmware, but check_for_dasharo_firmware will not detect any platform?
+        # According to check_for_dasharo_firmware it will return 1 in both
+        # cases which means that we cannot detect such case.
+        echo "Your credentials give access to:"
```
DaniilKl, line 1462: @EduKav1813, accidentally deleted your comment, sorry for that. Check comments above, they explain why all these echoes are inside condition.

include/dts-functions.sh
```diff
@@ -1453,19 +1453,34 @@ main_menu_options(){
...
+        echo -n "Dasharo Pro Package (DPP): "
+        print_ok "YES"
+      else
+        echo "Your credentials give access to:"
+	echo -n "Dasharo Pro Package (DPP): "
+        echo "NO"
+      fi
```
EduKav1813, line 1469: I think it would be better to first call the function, save its result in a variable, then print the common echo messages, and then have the if-else block.

DaniilKl: 1cdf6313557f03895b446d2dd8bbd477a02e7a84

EduKav1813 (approved):

PR #66, opened 2025-01-20
Provide solutions when printing warnings

PLangowski:


PLangowski:
@miczyg1 When ME is not disabled, we tell the user to use FlashBIOS https://docs.dasharo.com/guides/firmware-update/#known-issues.
What about `The firmware binary to be flashed contains Management Engine (ME), but ME is not writable!`? Is the solution identical here?

miczyg1:
> @miczyg1 When ME is not disabled, we tell the user to use FlashBIOS https://docs.dasharo.com/guides/firmware-update/#known-issues. What about `The firmware binary to be flashed contains Management Engine (ME), but ME is not writable!`? Is the solution identical here?

One can use FlashBIOS or simply use FUM and boot to DTS shell and flash the binary manually. The latter will work on any platform.

PLangowski:
@miczyg1 What are the exact steps that the user needs to take in order to take manual steps in DTS with FUM?

PLangowski:
@miczyg1 So for example sending the interrupt signal by pressing `CTRL+C`?

miczyg1:
> @miczyg1 So for example sending the interrupt signal by pressing `CTRL+C` will work?

No idea.

m-iwanicki:
@PLangowski you should maybe also rework ME warning as currently it's a little misleading. Issue: https://github.com/Dasharo/dts-scripts/issues/22 (should I migrate it to dasharo-issues?)

PLangowski:
I guess this has been implemented in https://github.com/Dasharo/dts-scripts/pull/71. Closing,

PR #67, opened 2025-01-22
Allow interrupting update in FUM

PLangowski:


m-iwanicki:
Fixed in https://github.com/Dasharo/dts-scripts/pull/114

PR #70, opened 2025-01-30
Hide DPP credentials by default

3mkusiak:
Displaying credentials risks accidental exposure. Hide credentials by default and add toggle option to display them.

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1332,9 +1332,15 @@ show_dpp_credentials() {
   if [ -n "${DPP_IS_LOGGED}" ]; then
     echo -e "${BLUE}**${NORMAL}                DPP credentials ${NORMAL}"
     echo -e "${BLUE}*********************************************************${NORMAL}"
-    echo -e "${BLUE}**${YELLOW}       Logs key: ${NORMAL}${CLOUDSEND_LOGS_URL}"
-    echo -e "${BLUE}**${YELLOW}   Download key: ${NORMAL}${CLOUDSEND_DOWNLOAD_URL}"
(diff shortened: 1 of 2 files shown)


PLangowski (approved):
Looks good. I tested it and it works as expected.

include/dts-functions.sh
```diff
@@ -1567,6 +1573,9 @@ show_footer(){
   else
     echo -e "${RED}${SEND_LOGS_OPT}${NORMAL} to enable sending DTS logs ${NORMAL}"
   fi
+  if [ -n "${DPP_IS_LOGGED}" ]; then
+    echo -e "${RED}${TOGGLE_DISP_CRED_OPT_UP}${NORMAL} to toggle displaying credentials ${NORMAL}"
```
m-iwanicki, line 1577: Maybe to `C to display/hide credentials` (display or hide would be displayed based on current state)?
Otherwise looks and works good.

3mkusiak: Ack, I'll change it.

3mkusiak: @m-iwanicki I changed it as you requested, please take a look.

m-iwanicki (approved):

include/dts-functions.sh
m-iwanicki: LGTM. When answering review comments please paste commit hash (or link) which fixes those requests.
In case of force push use `compare` link (before/after force push) to see only relevant changes

m-iwanicki (approved):

include/dts-functions.sh
3mkusiak: Ack

3mkusiak:
Already merged: https://github.com/Dasharo/dts-scripts/commit/60d87b5cb271dda45e2dc5a880e981f977b87c81

PR #71, opened 2025-02-04
Add `-N` (`--noverify-all`) if FD is locked

m-iwanicki:
Lately we had issues on MSI where some regions are not readable. Usual symptoms are locked FD and ME.
Somewhat related: https://github.com/Dasharo/dasharo-issues/issues/497
To make update possible we not only have to skip those parts but also read/verify only the regions we are flashing.

@miczyg1 I added you to reviewers. I'm mostly want to know if just skipping whole chip verification (`-N`) in case of locked FD is safe(ish) choice.

### DIFF
--- a/include/dts-functions.sh
+++ b/include/dts-functions.sh
@@ -1041,6 +1041,7 @@ set_flashrom_update_params() {
 }
 
 set_intel_regions_update_params() {
+  local fd_me_locked="no"
   if [ $BOARD_HAS_FD_REGION -eq 0 ]; then
     # No FD on board, so no further flashing
     FLASHROM_ADD_OPT_REGIONS=""
@@ -1055,6 +1056,7 @@ set_intel_regions_update_params() {
         # FD writable and the binary provides FD, safe to flash
         FLASHROM_ADD_OPT_REGIONS+=" -i fd"
       else
+        fd_me_locked="yes"
         print_error "The firmware binary to be flashed contains Flash Descriptor (FD), but FD is not writable!"
         print_warning "Proceeding without FD flashing, as it is not critical."
         echo "The firmware binary contains Flash Descriptor (FD), but FD is not writable!"  >> $ERR_LOG_FILE
@@ -1072,10 +1074,16 @@ set_intel_regions_update_params() {
           force_me_update
         fi
       else
+        fd_me_locked="yes"
         echo "The firmware binary to be flashed contains Management Engine (ME), but ME is not writable!"  >> $ERR_LOG_FILE
         print_error "The firmware binary contains Management Engine (ME), but ME is not writable!"
       fi
     fi
+  fi
+  if [ "$fd_me_locked" = "yes" ]; then
+    FLASHROM_ADD_OPT_REGIONS+=" -N"
+    print_warning "You can read more about issues with FD or ME on"
+    print_warning "https://docs.dasharo.com/guides/firmware-update/#known-issues"
   fi
 }
 


miczyg1:
> @miczyg1 I added you to reviewers. I'm mostly want to know if just skipping whole chip verification (`-N`) in case of locked FD is safe(ish) choice.

It is always best to verify if we accidentally didn't write to parts of flash we didn't intend to (no using an `-N`). However, in this case it is not possible and adding `-N` is the only way out.

include/dts-functions.sh
```diff
@@ -1067,16 +1070,22 @@ set_intel_regions_update_params() {
...
           force_me_update
         fi
       else
+        fd_me_locked="yes"
         echo "The firmware binary to be flashed contains Management Engine (ME), but ME is not writable!"  >> $ERR_LOG_FILE
         print_error "The firmware binary contains Management Engine (ME), but ME is not writable!"
       fi
     fi
   fi
+  if [ "$fd_me_locked" = "yes" ] && [[ "$SYSTEM_VENDOR" == *"Micro-Star International"* ]]; then
```
miczyg1, line 1085: I think we should not to limit this to MSI. There may appear new platform with similar problems in the future.

include/dts-functions.sh
```diff
@@ -1067,16 +1070,22 @@ set_intel_regions_update_params() {
...
+    print_warning "You can read more about issues with FD or ME on"
```
miczyg1, line 1086:
```suggestion
    FLASHROM_ADD_OPT_REGIONS+=" -N"
    print_warning "You can read more about issues with FD or ME on"
```

It will be better to move it here, because whether ME is locked or FD is locked, we must use `-N`. Sometimes ME is not readable and will cause the same error as the hole between ME and BIOS regions.

include/dts-functions.sh
```diff
@@ -1055,6 +1056,8 @@ set_intel_regions_update_params() {
         # FD writable and the binary provides FD, safe to flash
         FLASHROM_ADD_OPT_REGIONS+=" -i fd"
       else
+        FLASHROM_ADD_OPT_REGIONS+=" -N"
```
miczyg1, line 1059:
```suggestion
```

include/dts-functions.sh
```diff
@@ -1067,16 +1070,22 @@ set_intel_regions_update_params() {
         if [ $ME_DISABLED -eq 1 ]; then
           FLASHROM_ADD_OPT_REGIONS+=" -i me"
         else
+          fd_me_locked="yes"
```
miczyg1, line 1073: This case doesn't mean ME locked, but not disabled. That's different from being locked.

miczyg1 (changes requested):

include/dts-functions.sh
m-iwanicki: 306dd78dbfbb

include/dts-functions.sh
m-iwanicki: 306dd78dbfbb

include/dts-functions.sh
m-iwanicki: 306dd78dbfbb

m-iwanicki:
@miczyg1 answered your review: https://github.com/Dasharo/dts-scripts/compare/b641072c3b3bc953d876882b2bd6144b0c1e106e..306dd78dbfbb74068da1d0ed35ae9025b78ba9de

miczyg1 (approved):

PR #72, opened 2025-02-04
scripts: fix double ME flashing dialog

3mkusiak:
During attempt to update Dasharo via DTS user is asked twice to skip ME flashing. Move `set_intel_regions_update_params()` call from `firmware_pre_updating_routine()` to the "else" part of the "if" statement after `firmware_pre_updating_routine()` call so it is not executed twice.

### DIFF
--- a/scripts/dasharo-deploy
+++ b/scripts/dasharo-deploy
@@ -749,7 +749,6 @@ firmware_pre_updating_routine(){
   check_blobs_in_binary $BIOS_UPDATE_FILE
   check_if_me_disabled
   set_flashrom_update_params $BIOS_UPDATE_FILE
-  set_intel_regions_update_params "-N --ifd"
   check_vboot_keys
 
   return 0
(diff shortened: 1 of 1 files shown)


3mkusiak:
Needs to be tested on bare metal.

3mkusiak:
I confirm the fix is working or bare metal.

m-iwanicki (approved):

macpijan:
@artur-rs Can we ensure this  check will go into automated test scenario?

artur-rs:
@macpijan sure, adding to the test-scope backlog

EduKav1813:
@3mkusiak Could you please share the steps needed to reproduce the original problem in DTS?

3mkusiak:
Already merged: https://github.com/Dasharo/dts-scripts/commit/099090b5b04e46de930c43d628a63563ff272074

PR #73, opened 2025-02-05
dts-functions.sh: fix odroid link and add CAN_INSTALL_BIOS var

m-iwanicki:


PLangowski (approved):

include/dts-functions.sh
```diff
@@ -633,7 +634,7 @@ board_config() {
           ;;
       esac
 
-      BIOS_LINK_DPP="$FW_STORE_URL_DPP/$DASHARO_REL_NAME/v$DASHARO_REL_VER_DPP/${DASHARO_REL_NAME}_v$DASHARO_REL_VER_DPP.rom"
+      BIOS_LINK_DPP="$FW_STORE_URL_DPP/odroid_h4/v$DASHARO_REL_VER_DPP/${DASHARO_REL_NAME}_v$DASHARO_REL_VER_DPP.rom"
```
DaniilKl, line 637: @m-iwanicki, why not to change `DASHARO_REL_NAME` for `ODROID-H4`? On the other hand, why you assign `*/odroid_h4` to all `HARDKERNEL` instead of only to `ODROID-H4`? It clould be:

```bash
BIOS_LINK_DPP="$FW_STORE_URL_DPP/${DASHARO_REL_NAME}/v$DASHARO_REL_VER_DPP/hardkernel_${DASHARO_REL_NAME}_v$DASHARO_REL_VER_DPP.rom"
```

m-iwanicki: Yes, that would be better. https://github.com/Dasharo/dts-scripts/pull/75.
I don't think we need to add this change to current release as we only have one odroid.

PR #77, opened 2025-03-06
install_workflow: Exit when trying to flash v1.1.1 on 13th gen and above

PLangowski:


scripts/dasharo-deploy
```diff
@@ -917,6 +917,23 @@ install_workflow() {
...
     verify_artifacts bios
   fi
+  
+  # TODO: It is not a good practice to do some target specific work in the code
+  # of a scallable product, this should be handled in a more scallable way:
+  if [[ "$UPDATE_VERSION" == "1.1.1" && \
+    ( "$BOARD_MODEL" == "PRO Z690-A WIFI DDR4(MS-7D25)" || \
+      "$BOARD_MODEL" == "PRO Z690-A DDR4(MS-7D25)" || \
+      "$BOARD_MODEL" == "PRO Z690-A (MS-7D25)" || \
+      "$BOARD_MODEL" == "PRO Z690-A WIFI (MS-7D25)" ) ]]; then
```
m-iwanicki, line 927: Maybe replace with `"$SYSTEM_MODEL" = "MS-7D25"`?

scripts/dasharo-deploy
```diff
@@ -917,6 +917,23 @@ install_workflow() {
     download_bios
     verify_artifacts bios
   fi
+  
+  # TODO: It is not a good practice to do some target specific work in the code
+  # of a scallable product, this should be handled in a more scallable way:
```
m-iwanicki, line 922: Maybe

```sh
do_nothing() {
    :
}

min_intel_cpu_gen() {
    local min_gen=$1
    local workflow=$2
    if [ "$(get_intel_gen)" -lt "$min_gen" ]; then
        return 1;
    else
        return 0;
    fi
}

# use: '$PRE_DEPLOY_CHECK <deploy_workflow>'
PRE_DEPLOY_CHECK=do_nothing
(...)
case "$SYSTEM_MODEL" in
        "MS-7D25")
            PRE_DEPLOY_CHECK="min_intel_cpu_gen 11"
            (...)

(...)
install() {
    $PRE_DEPLOY_CHECK install
}
```

That way we could create different checks for different platforms if needed

PR #83, opened 2025-04-26
scripts/dasharo-deploy.sh: Fix including ACMs in binaries

miczyg1:
BIOS ACM must be properly aligned in the flash image, typically 128K, depends on ACM size.

Compress SINIT ACM to save space.


SergiiDmytruk (approved):
Didn't use the script, but using those parameters manually does work fine on hardware.

PR #84, opened 2025-06-02
Use dts_config repo

PLangowski:


include/dts-functions.sh
```diff
@@ -230,6 +230,8 @@ board_config() {
 
   wait_for_network_connection
 
+  git clone https://github.com/Dasharo/dts-configs.git "$BOARD_CONFIG_PATH"
+  (cd "$BOARD_CONFIG_PATH" && git checkout add-configs)
```
DaniilKl, line 234: Why the `checkout`? I suppose all configurations that will be used in the release image will be on `master` (or `main`).

include/dts-functions.sh
```diff
@@ -230,6 +230,8 @@ board_config() {
 
   wait_for_network_connection
 
+  git clone https://github.com/Dasharo/dts-configs.git "$BOARD_CONFIG_PATH"
```
DaniilKl, line 233: You could redirect the output to logs and add echos informing the user what is going on here.

include/dts-functions.sh
```diff
@@ -1845,3 +1808,29 @@ ask_for_confirmation() {
...
+  # If separate BOARD_MODEL values exist, parse the variables 
+  eval $(jq -r --arg m $(echo "$system_model" | tr '[:upper:]' '[:lower:]') --arg b $(echo "$board_model" | tr '[:upper:]' '[:lower:]') '
+  if .models[$m].board_models[$b] then
+    .models[$m].board_models[$b]
+    | to_entries[]
+    | "\(.key | ascii_upcase)=\"\(.value|tostring)\""
+  else
+    empty
+  end
+' $json_file)
```
DaniilKl, line 1835: How do you not modifying the default values set in `dts-environment.sh` in case the variables are not set in JSON's? I think you should describe in details what you are doing here and how you are doing it.

DaniilKl (changes requested):

include/dts-functions.sh
PLangowski: Yes, this will be changed after the `add-configs` branch is merged into main

include/dts-functions.sh
PLangowski: I'm not sure what you mean by this question.
This function loads the json (`<vendor>.json`) and parses selected fields into variables of the same names (uppercase). By `selected` I mean the following:
- All top-level key-value pairs (common for the board vendor)
- key-value pairs found in `models[$SYSTEM_MODEL]` (model-specific values)
- If the JSON has different `$BOARD_MODEL` values, it also parses those (for now this is not used, but it will be for some platforms)

include/dts-functions.sh
```diff
@@ -230,6 +230,13 @@ board_config() {
 
   wait_for_network_connection
 
+  echo "Cloning board configs repository to $BOARD_CONFIG_PATH"
+  git clone https://github.com/Dasharo/dts-configs.git "$BOARD_CONFIG_PATH" >/dev/null 2>>"$ERR_LOG_FILE"
```
macpijan, line 234: Can we have it already cloned, e.g. when running off USB?

include/dts-functions.sh
```diff
@@ -1845,3 +1813,29 @@ ask_for_confirmation() {
     esac
   done
 }
+
+parse_config() {
+  local vendor="$1"
+  local system_model="$2"
+  local board_model="$3"
+  json_file="$BOARD_CONFIG_PATH/configs/$(echo "$vendor" | tr '[:upper:]' '[:lower:]' | sed 's/ /_/g').json"
```
macpijan, line 1821: It does not have to block this PR, but my impression is: if we already have python anyway and use it in dts, I would personally write a simple python3 script and execute it there.

I wonder if one can easily understand this jq code without a headache.

include/dts-functions.sh
macpijan: Why do we need git in the first place then, rather than downloading simply the latest config tarball from github?

We wouldn't need to add git to the image also.

macpijan: Since we are reinventing fwupd here already, I would advise to check out how fwupd refreshes metadata. It also downloads tarball.

include/dts-functions.sh
DaniilKl: No reason to make it build in, because we still need to download the firmware and hence have access to network?

include/dts-functions.sh
DaniilKl: Please add comments describing what exactly you are doing here. Right now this code is not readable enough.

include/dts-functions.sh
PLangowski: @macpijan It's better to update this every time we run Dasharo update/install. That way we can bump firmware version without having to publish a new DTS release

include/dts-functions.sh
PLangowski: https://github.com/Dasharo/dts-scripts/compare/5af8b57b551458920ce7e7eb67eedac837678fbc..ce24f25db6269edf1e2e1baa8836f83092201fde

include/dts-functions.sh
PLangowski: A python script executed as a subprocess of the bash script would not be able to modify variables that easily

DaniilKl: Ok. Please explain what you are doing here in details.

PLangowski: 6252bbf007258f013aae05ee480305cef3be04c1

DaniilKl: The lines:

```
  # This finds all keys in `models[$SYSTEM_MODEL]` other than `board_models`Add commentMore actions
  # and assigns variables just like in the previous call
```

Say the same as the line:

```
  # Parse system model-specific variables
```

Same for other comments you have added. Could you describe every tool call and every argument in these lines?

PLangowski: I don't see why that would be necessary. If you read the comments and look at the commands again, they become pretty clear

DaniilKl:
```suggestion
parse_config() {
  local vendor="$1"
  local system_model="$2"
  local board_model="$3"
  # The JSONs names in configs directory the same as values in vendor variable but lowercase and spaces replaced with _
  json_file="$BOARD_CONFIG_PATH/configs/$(echo "$vendor" | tr '[:upper:]' '[:lower:]' | sed 's/ /_/g').json"
  # This finds all keys other than `models` and maps them to bash variables.
  # to_entries[] | select(.key != "models")
  # Converts JSON input from:
  # "key1": "value"
  # "key2": "value"
  # "models": {
  #   "key3": "value"
  #  }
  # To:
  # "key1": "value"
  # "key2": "value"
  #
  # "\(.key | ascii_upcase)=\"\(.value|tostring)\"
  # Makes from: "key1": "value" - KEY1="value"
  eval $(jq -r 'to_entries[] | select(.key != "models") | "\(.key | ascii_upcase)=\"\(.value|tostring)\""' $json_file)
  # Parse system model-specific variables
  # This finds all keys in `models[$SYSTEM_MODEL]` other than `board_models`
  # and assigns variables just like in the previous call
  eval $(jq -r --arg m $(echo "$system_model" | tr '[:upper:]' '[:lower:]') '
  .models[$m] 
  | to_entries[] 
  | select(.key != "board_models")
  | "\(.key | ascii_upcase)=\"\(.value|tostring)\""
' $json_file)
  # If separate BOARD_MODEL values exist, parse the variables 
  # This looks for `models[$SYSTEM_MODEL].board_models[$BOARD_MODEL]`. Some 
  # boards have different variable values for different board models.
  eval $(jq -r --arg m $(echo "$system_model" | tr '[:upper:]' '[:lower:]') --arg b $(echo "$board_model" | tr '[:upper:]' '[:lower:]') '
  if .models[$m].board_models[$b] then
    .models[$m].board_models[$b]
    | to_entries[]
    | "\(.key | ascii_upcase)=\"\(.value|tostring)\""
  else
    empty
  end
' $json_file)
```

PLangowski: Alright, I added that and some more
https://github.com/Dasharo/dts-scripts/compare/6252bbf007258f013aae05ee480305cef3be04c1..f1a7cd624df766182a7457d77ed921d034348822

PLangowski: https://github.com/Dasharo/dts-scripts/compare/f1a7cd624df766182a7457d77ed921d034348822..bf0736f691b29e0c3c5fc9a34e04c248e26efd81

include/dts-functions.sh
```diff
@@ -301,12 +310,7 @@ board_config() {
       fi
       ;;
     "NV4xPZ")
-      DASHARO_REL_NAME="novacustom_nv4x_adl"
-      DASHARO_REL_VER="1.7.2"
-      HEADS_REL_VER_DPP="0.9.1"
-      HEADS_LINK_DPP="${BUCKET_DPP_HEADS}/${DASHARO_REL_NAME}/v${HEADS_REL_VER_DPP}/${DASHARO_REL_NAME}_v${HEADS_REL_VER_DPP}_heads.rom"
```
DaniilKl, line 307: Where do you set `HEADS_LINK_DPP` for this platform?

DaniilKl (changes requested):

include/dts-functions.sh
```diff
@@ -595,52 +599,16 @@ board_config() {
...
-    "APU4")
-      DASHARO_REL_NAME="pcengines_apu4"
-      ;;
-    "APU6")
-      DASHARO_REL_NAME="pcengines_apu6"
-      ;;
-    *)
-      print_error "Board model $SYSTEM_MODEL is currently not supported"
-      return 1
-      ;;
```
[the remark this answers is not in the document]
DaniilKl: Maybe you should make `parse_config` return `1` if it did not found needed configuration?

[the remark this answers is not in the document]
PLangowski: As of now, jq does not fail when it doesn't find the right key. It would require rewriting the function. I think for now we can leave it like this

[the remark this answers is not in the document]
DaniilKl: Perfect, thank you

include/dts-functions.sh
```diff
@@ -595,53 +611,36 @@ board_config() {
...
+      print_error "Vendor $VENDOR is currently not supported!"
+      return 1
+    elif [ $? -eq 2 ]; then
+      print_error "System model $SYSTEM_MODEL is currently not supported!"
+      return 1
+    elif [ $? -eq 3 ]; then
+      print_error "Board model $BOARD_MODEL is currently now supported"
+      return 1
+    fi
+;;
```
DaniilKl, line 643:
```suggestion
      parse_config "$SYSTEM_VENDOR" "$SYSTEM_MODEL" "$BOARD_MODEL"
      if [ $? -eq 1 ]; then
        print_error "Vendor $VENDOR is currently not supported!"
        return 1
      elif [ $? -eq 2 ]; then
        print_error "System model $SYSTEM_MODEL is currently not supported!"
        return 1
      elif [ $? -eq 3 ]; then
        print_error "Board model $BOARD_MODEL is currently now supported"
        return 1
      fi
      ;;
```

PLangowski: https://github.com/Dasharo/dts-scripts/compare/0ebe889238a9a84fcb3ecd02b83498d407decaff..c70a0c8bfb6e4eb6c64d9f647d237b255124ab65

DaniilKl (approved):

include/dts-functions.sh
```diff
@@ -230,6 +230,15 @@ board_config() {
 
   wait_for_network_connection
 
+  echo "Downloading board configs repository to $BOARD_CONFIG_PATH.tar.gz"
+  mkdir -p "$BOARD_CONFIG_PATH"
+  curl -L -o "$BOARD_CONFIG_PATH.tar.gz" https://github.com/Dasharo/dts-configs/archive/refs/heads/add-configs.tar.gz >/dev/null 2>>"$ERR_LOG_FILE"
```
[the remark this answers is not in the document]
PLangowski: https://github.com/Dasharo/dts-scripts/pull/84/commits/ea45864477f34977976f8bdef6d527aee9c05e55

DaniilKl (approved):

PR #87, opened 2025-06-10
Add pcengines seabios update workflow

m-iwanicki:
As of now I assume every pcengines fw is Dasharo.
To check if we want to offer SeaBIOS update I check if coreboot config contains `CONFIG_PAYLOAD_SEABIOS=y`. If it does then fw is SeaBIOS and we can update.
In case of UEFI coreboot shouldn't have this string and should offer UEFI update. Unfortunately as of now there is no way to transition to UEFI from SeaBIOS and as there is no initial deployment, transition would have to be done manually.


scripts/dasharo-deploy.sh
```diff
@@ -245,6 +245,33 @@ choose_version() {
     fi
   fi
 
+  if [ -n "$DASHARO_REL_VER_DPP_SEABIOS" ]; then
+    tmp_rom=$(mktemp --dry-run)
+    config=/tmp/config
+    # get current firmware
+    for flash_name in $FLASH_CHIP_LIST; do
```
PLangowski, line 252: I might not understand this correctly, but what if there are several chips with different firmware and one of them is not seabios?

m-iwanicki: I could probably replace whole for with `$FLASH_CHIP_SELECT` as this function does pretty much the same thing (loops over all possible chips to find one we can use): https://github.com/Dasharo/dts-scripts/blob/04c62930f36e5d5e84e5861cc6ea015a82707a58/include/dts-functions.sh#L720

miczyg1: > I might not understand this correctly, but what if there are several chips with different firmware and one of them is not seabios?

it is not currently possible to have different firmwares on one platform. Although it is possible to have a platform with different hardware BOMs (e.g. different flash chips). This loop handles this case where there is a possibility to have different flash chip model. MSI desktops are such a case, and maybe others.

include/dts-functions.sh
```diff
@@ -1399,6 +1409,13 @@ show_main_menu() {
   if [ -f "${DPP_SUBMENU_JSON}" ]; then
     echo -e "${BLUE}**${YELLOW}     ${DPP_SUBMENU_OPT})${BLUE} DTS extensions${NORMAL}"
   fi
+  # As of now show this option only for PC Engines and only for non-UEFI
+  # firmware as we only implement transition to UEFI for PC Engines
+  # TODO: migrate all transition logic here e.g. Heads, UEFI->SeaBIOS if
+  # possible
+  if [[ ! -d "/sys/firmware/efi" && "$SYSTEM_VENDOR" == "PC Engines" ]]; then
```
DaniilKl, line 1416: This should be mocked with DTS HAL :thinking:

m-iwanicki: If only we had some easy to read documentation so I don't need to dig through code to find out what is or isn't implemented in HAL.

DaniilKl, line 1416:
```suggestion
  if [[ ! $(fsread_tool_test_mock -d "/sys/firmware/efi") && "$SYSTEM_VENDOR" == "PC Engines" ]]; then
```

DaniilKl:
```suggestion
  if [[ ! $(fsread_tool_test_mock -d "/sys/firmware/efi") && "$SYSTEM_VENDOR" == "PC Engines" ]]; then
```

m-iwanicki: e920a022260d83f4d2e1027d9a88a643fec1e024

include/dts-functions.sh
```diff
@@ -1559,6 +1576,16 @@ main_menu_options() {
   "${DPP_SUBMENU_OPT}")
     [ -f "$DPP_SUBMENU_JSON" ] || return 0
     export DPP_SUBMENU_ACTIVE="true"
+    return 0
+    ;;
+  "${TRANSITION_OPT}")
+    ${CMD_DASHARO_DEPLOY} transition
+    result=$?
+    if [ "$result" -ne $OK ] && [ "$result" -ne $CANCEL ]; then
+      send_dts_logs ask && return 0
```
DaniilKl, line 1585:
```suggestion
      send_dts_logs ask && return $OK
```

scripts/dasharo-deploy.sh
```diff
@@ -1088,6 +1111,169 @@ update_workflow() {
...
+    d | D | dpp | DPP | Dpp)
+      if [ -n "$_might_be_dpp" ]; then
+        print_ok "Subscription version (cooreboot + EDK2) selected"
+        FIRMWARE_VERSION="dpp"
+        break
+      fi
+      ;;
+    b | B)
+      echo "Returning to main menu..."
+      exit 0
```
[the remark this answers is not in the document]
DaniilKl: Or maybe "$CANCEL" :thinking:

scripts/dasharo-deploy.sh
```diff
@@ -1088,6 +1111,169 @@ update_workflow() {
...
+  firmware_pre_installation_routine
+
+  echo "Transitioning Dasharo firmware..."
+  # FIXME: It seems we do not have an easy way to add some flasrhom extra args
+  # globally for specific platform and variant
+  $FLASHROM -p "$PROGRAMMER_BIOS" ${FLASH_CHIP_SELECT} ${FLASHROM_ADD_OPT_REGIONS} -w "$BIOS_UPDATE_FILE" >>$FLASHROM_LOG_FILE 2>>"$ERR_LOG_FILE"
+  error_check "Failed to transition Dasharo firmware"
+  print_ok "Successfully transitioned Dasharo firmware"
+
+  return 0
```
DaniilKl, line 1230:
```suggestion
  return $OK
```

PR #102, opened 2025-08-28
Prevent deploying Dasharo v1.1.1 on 13th gen and newer CPUs on incompatible board models

aronowski:
Prevent deploying Dasharo v1.1.1 on 13th gen and newer CPUs on incompatible board models both in the update workflow and the install workflow.

Resolves https://github.com/Dasharo/dasharo-issues/issues/1261

Currently this has been tested manually, by mocking the relevant platforms on QEMU. The relevant OSFV tests will be implemented as per https://github.com/Dasharo/open-source-firmware-validation/pull/1045.


scripts/dasharo-deploy.sh
```diff
@@ -963,6 +963,25 @@ deploy_firmware() {
   return 1
 }
 
+prevent_deploying_1_1_1_on_13th_gen_and_newer() {
```
m-iwanicki, line 966: I don't think that's a good function name.
Maybe use some kind of generic `check_if_cpu_compatible` and mention in function documentation that as of now it only checks MSI.
That way we could expand it in the future if needed.

m-iwanicki (changes requested):

include/hal/common-mock-func.sh
```diff
@@ -527,12 +527,13 @@ setpci_check_me_op_mode_mock() {
 ################################################################################
 # lscpu
 ################################################################################
-TEST_CPU_MODEL="${TEST_CPU_MODEL:-test}"
+TEST_CPU_VERSION="${TEST_CPU_VERSION:-test}"
```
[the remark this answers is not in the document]
aronowski: fixed: https://github.com/Dasharo/dts-scripts/compare/dafd80549a3776dcd73f454a257b6c67903199d2..f6126f6920f327dfc614c21b2d06fe902e1914d7#diff-ffa8fd4664ec7ece80472f9a0136f2eb2a6e4e0ba8217a3017739c2321ae153dL530

scripts/dasharo-deploy.sh
```diff
@@ -963,6 +963,32 @@ deploy_firmware() {
...
+  # details.
+  # TODO: It is not a good practice to do some target specific work in the code
+  # of a scallable product, this should be handled in a more scallable way:
+  if [[ "$UPDATE_VERSION" == "1.1.1" &&
+    ("$BOARD_MODEL" == "PRO Z690-A WIFI DDR4(MS-7D25)" ||
+    "$BOARD_MODEL" == "PRO Z690-A DDR4(MS-7D25)" ||
+    "$BOARD_MODEL" == "PRO Z690-A (MS-7D25)" ||
+    "$BOARD_MODEL" == "PRO Z690-A WIFI (MS-7D25)") ]]; then
+
+    cpu_gen_check=$(echo "$CPU_VERSION" | grep -E "\-(13|14)[0-9]{3}" | wc -l)
```
m-iwanicki, line 982: I'm not sure about this part: <https://en.wikipedia.org/wiki/Raptor_Lake#Raptor_Lake-S_Refresh>

Can you put `300` or `300T` in MSI?

aronowski: Might extend the regex to `\-(13|14)[0-9]{3}| 300` just to be safe and document it, linking the relevant Intel specification pages.

aronowski: fixed: https://github.com/Dasharo/dts-scripts/compare/88df835db3f30570f3fcf23ba95483cf3c66f88a..23068c2abaacc3590ec40ca8642547bc6290f0f6

albeit with the wiki entry, not Intel specification pages.

m-iwanicki: Will it work with space? `| 300`? maybe:

```diff
-| 300
+|300
```

PR #115, opened 2025-09-29
Add FUM, smmstore and flash layout mocking

m-iwanicki:
Related to changes and fixes done in v2.7.1


3mkusiak (approved):
LGTM

Tested this manually, tried various combinations to break it but couldn't.
Two things I noticed though:
* Emulating FUM does not prompt for TUI refresh, instead it just gives shell control.
* Tried FUM for MSI initially, and got prompt to connect power brick. Though, this seems not to be related at all with your changes. Just wanted to point it out.

Will test this automatically under PR on osfv.

PR #116, opened 2025-10-01
bash-ui-lib: testing claude

macpijan:


bash-ui-lib/lib/ui-colors.sh
DaniilKl: This lib looks ok.

bash-ui-lib/lib/ui-render.sh
```diff
@@ -0,0 +1,363 @@
...
+  echo -ne "${UI_THEME[PROMPT]}\nEnter an option:${UI_COLORS[NORMAL]} "
+}
+
+# ui_render_footer
+# Render footer (uses callback if set, otherwise default)
+ui_render_footer() {
+  if [[ -n "$UI_FOOTER_CALLBACK" ]] && type -t "$UI_FOOTER_CALLBACK" &>/dev/null; then
+    "$UI_FOOTER_CALLBACK"
+  else
+    ui_render_default_footer
```
DaniilKl: A quick note. I do not like that we delegate drawing footer to some other function. It looks like the library does not know what structure should be used exactly and how to draw it. We should make sure that the library knows that we want to use the following footer:

image

bash-ui-lib/lib/ui-core.sh
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/env bash
+
```
DaniilKl: Currently we have 3 types of output that a user can see:
1. The output we want the user to see.
2. The output we do not want the user to see.
3. The output we want user to see when a certain condition is met (e.g. the debug logs should be seen only when turned on).

The third one might be questionable.

The example of a command that produces the `1` output:

```
echo -ne "${UI_THEME[PROMPT]}\nEnter an option:${UI_COLORS[NORMAL]} "
```

The example of a command that produces the `2` output:

```
cbfstool <args>
```

Both these commands are printing their `STDOUT` (and `STDERR` as well) to the default file descriptor defined by the `console` Linux kernel command line parameter.

In DTS we typically redirect the `2` output to `/dev/null`. But there are cases when we do not, for example if we forget or some tool will unexpectedly start to throw smth. to `STDOUT`. These cases appear in the backends and are not controlled by frontend, hence they could ruin the UI experience.

The idea is to print all the output from the UI libraries not to the file descriptor defined by the `console` Linux kernel command line parameter, but to some other. And then display the UI from this descriptor. This way we will not care whether some tool unexpectedly throus smth to the file descriptor defined by the `console` Linux kernel command line parameter. But this may bring up some TTY redirection issues.

bash-ui-lib/lib/ui-core.sh
```diff
@@ -0,0 +1,329 @@
...
+# SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# ui-core.sh - Core UI engine and main loop for bash-ui-lib
+
```
DaniilKl: Currently we have all global (not all but many actually) DTS variables in one bash env.. Bringing another library with the configuration stored in global bash variables will mess upp the environment further. For example some backend can overwrite the UI configuration variable without calling to the specific UI function, hence messing up UI configuration.

The idea is to make the UI not a simple library that will consists of the functions that can be called from backend like:

```bash
source ui-render.sh

ui_print_info "Hello world!"
```

Where the `ui_render_foote` uses global variables that are visible by the backend code as well. But to make the UI a bash script (or a set of bash scripts) twith a small library consisting of simple functions that will:
1. Provide the ability of modifying UI static configuration that is stored, for example, in YAML file.
2. Provide the ability to communicate the dynamic variables via a defined set of variables (this set of variables already exists in DTS) to the UI. For example:

```bash
source ui-helper-functions.sh

# Function that changes static configuration in YAML
configure_print_color yellow


declare -g info
info="Hello world!"
# The function that redirects some message via global variable without saving it to YAML or using sockets or any other additional technologies by providing the name of the dynamic variable to the UI script.
ui_command.sh ui_print_info info

**Review by DaniilKl — COMMENTED** [review]
I have reviewed only code that is important from the architecture point of view.

File: bash-ui-lib/lib/ui-core.sh (hunk 2 above)
Comment (DaniilKl) [reply bash-ui-lib/lib/ui-core.sh]: This is a @m-iwanicki idea.

File: bash-ui-lib/lib/ui-core.sh (hunk 3 above)
Comment (DaniilKl) [reply bash-ui-lib/lib/ui-core.sh]: This is a @m-iwanicki idea.

File: bash-ui-lib/lib/ui-core.sh [hunk 4]
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
macpijan: something like this 28fc9ea9dfdfb18ed177e93a171c91ffb495da3e ?

PR #123, opened 2025-11-17
preventing message duplication

AsiaEwa:
preventing message duplication: "Waiting for network connection " only once
The modified dts-functions.sh file contains my proposed solution (the same name, but placed in main to distinguish it; the base file remains in "include"):
The first time the function is called, the message will be displayed. Subsequent calls will no longer show it until the script finishes or the flag is manually reset (NETWORK_WAIT_MSG_SHOWN=false).

### DIFF
--- a/dts-functions.sh
+++ b/dts-functions.sh
@@ -1,0 +1,1706 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# shellcheck disable=SC2034
+
+### Color functions:
+function echo_green() {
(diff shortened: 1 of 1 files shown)


macpijan:
> The modified dts-functions.sh file contains my proposed solution (the same name, but placed in main to distinguish it; the base file remains in "include"):

I am afraid this is not how you change stuff to be reviewed.

The proper way would be to actually edit the target file. Otherwise, it's not possible to say which changes are yours, and which not. Especially since there is only one commit here.

Take a look at: https://github.com/Dasharo/dts-scripts/pull/123/files

Now I would need to manually diff the original file and yours to have a grasp of what has changed. This is unlikely to get a review in practice, since so much effort is needed on the reviewer side.

PR #128, opened 2025-12-05
scripts/btg_key_validator.sh: add script for verifying the current BtG key

mkopec:


miczyg1:
TODOs for later to make it more generic:

1. Grep for SHA length in the km-show output to use proper hashing algorithm to obtain the key hash.
2. Reuse the DTS env variables o identify the platform and the key which the validator should check against.
3. Make the script take 2 arguments: file path and expected hash. This would make the script usable to verify binaries offline before flashing. Of course if parameters are given, flashrom step should be omitted and key hash should be taken from second parameter.
4. Instead of printing hardcoded messages about error, we should just print generic message that the key n the binary does not match expected hash. To be decided if we print additional info like filename/platform name (depending on use case described in point 3) and keyhash vs expected.

DaniilKl (changes requested):

scripts/btg_key_validator
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Extracting key manifest..."
+bg-prov km-export $rom km.bin >/dev/null
+
+modulus=$(bg-prov km-show km.bin | grep "Key And Signature" -A 8 | grep Data | cut -d ' ' -f 10 | tail -c +8)
+exponent=01000100
+
+echo $modulus$exponent | awk '{gsub(/.{2}/,"& ")}1' | xxd -r -p | sha384sum | grep -q $ncm_mtl_key
+
+if [ $? -ne 0 ]; then
```
miczyg1:
```suggestion
if [ $? -eq 0 ]; then
```

miczyg1 (approved):

PR #151, opened 2026-03-18
include: dts-functions.sh: sync_clocks before dumping and sending HCL

DaniilKl:
If clocks are not synced the "mc" might fail with the following log:

```
mc: <ERROR> Unable to initialize new alias from the provided credentials. The difference between the request time and the server'\''s time is too large.
```

And the HCL will not be sent.


ghost (approved):
LGTM, merge when ready

